PR #59, opened 2012-09-10
Canada Post Platform Web Service

jnormore:
This pull request contains integration for the new Canada Post web services for onboarding, calculated rates, tracking, and label printing on behalf of a merchant. Details about their new services are at 

http://www.canadapost.ca/cpotools/apps/developers/home?execution=e1s1

For each actions (oboarding, rates, tracking, and label/shipment creation, service methods and options retrieval) there is a method to initiate the action, a method to build the xml request, and a method to parse the xml response.

I've added a new PackageItem class to facilitate a list of items (with skus, hs codes, etc) within a package in the label/shipment creation methods.

Onboarding is accomplished by retrieving a request token, and redirecting user to CP login/signup page with this token, and then listening for a response from CP containing this token and a success flag.

Rating is similar to rate retrieval in other services, it requires a set of shipping options, list of items, etc.. and a list of available shipping services and rates are returned.

Tracking expects a tracking number and a response containing a list of tracking events (location, status, message) are returned.

Label/shipment retrieval is a 2 step process, first the shipment is created with a set of shipping options, line items (containing skus, hs codes, weights, etc), and package details (dimensions, weight), and the resulting shipping_id is used to retrieve the actual printed label (pdf file).

Shipping service methods and available options retrieval is also added. This is useful for a client to display available services and options for those services (where options can depend on each other either as prerequisites or conflicts) since the rating and shipment services are very strict on requiring these option dependencies in requests.

Besides unit tests, remote tests are added for both acting on behalf of merchants and not acting on behalf of merchants. Acting on behalf of merchants is required when creating labels, but not required for the other services.

@jduff  and @odorcicd Can you review please, Thanks!


megetron:
@jnormore, 
trying to use this implementation without success, can you please provide a basic usage that does work (like the ones you have on active_shipping github page)?


I was looking for examples on remote/test but couldn't make it work.
request:
data: 
    "<?xml version=\"1.0\"?>\n<mailing-scenario xmlns=\"http://www.canadapost.ca/ws/ship/rate\">\n  <customer-number>0001234567</customer-number>\n  <quote-type>commercial</quote-type>\n  <parcel-characteristics>\n    <weight>0.001</weight>\n    <mailing-tube>false</mailing-tube>\n    <unpackaged>false</unpackaged>\n  </parcel-characteristics>\n  <origin-postal-code>37862</origin-postal-code>\n  <destination>\n    <united-states>\n      <zip-code>10001</zip-code>\n    </united-states>\n  </destination>\n</mailing-scenario>\n"
headers: {"Authorization"=>"Basic some_code\nsome_code2==\n", "Accept-Language"=>"en-CA", "Accept"=>"application/vnd.cpc.ship.rate+xml", "Content-Type"=>"application/vnd.cpc.ship.rate+xml"}

The following end points (not sure what the differences between both):
endpoint1: "https://ct.soa-gw.canadapost.ca/rs/ship/price"
endpoint2: "https://soa-gw.canadapost.ca/rs/ship/price "


I can't get response XML probably because the response is Bad Request (400 - page not found)
here is the code to generate this:
    canada_post = ActiveShipping::CanadaPostPWS.new(customer_number: "123456",  api_key: "my_login_id : my_password", secret: "my_password" ,:test => true)
response = shipper.find_rates(origin, destination, packages)

PR #150, opened 2014-02-19
USPS address validation

smartacus:
Here's a fairly basic implementation of USPS address validation. I have most of the code for FedEx validation as well, but I need to clean it up a bit before submitting a pull request.


wvanbergen:
Sorry, but this needs rebasing again. The USPS carrier implementation was rewritten to use Nokogiri instead of REXML / XmlNode

PR #161, opened 2014-06-04
Added ability to find all tracking numbers with shipment id

alexgirdler:
Finds and returns all tracking numbers associated with a shipment ID as well as returning tracking_response for each tracking number.
Due to the double UPS API call in find_all_tracking_info I was unable to create a unit test.


wvanbergen:
When would you use this call? Is this UPS specific?

- Is it possible to add a remote test, that tests both API calls in succession?
- This will need to be rebased on the latest master, because the UPS carrier was rewritten using Nokogiri.

alexgirdler:
Sorry I did not see that you commented on this. From what I have seen it is UPS specific. The situation this was used in was as follows:

Order bulk merchandise from vendor.
Vendor provides only partial number of tracking numbers(I believe we were missing 3 unique tracking numbers)
When packages actually arrive there will be extra tracking numbers not included in shipment information that are linked through UPS to main parcel(bulk orders for UPS all share a common parent)

I will rebase and push. As for the remote tests I am not sure if I would be able to I will look into writing one and get back with you.

PR #203, opened 2014-11-26
UPS LTL Rate and Ship

slatem:
Created a new class to rate and ship UPS LTL shipments using the UPS LTL Rating Soap API. Currently there are 6 remote unit tests. Options for lift gate, and residential pickup. Need to add the shipper_number to your ups fixture. It does add the nokogiri gem which may be frowned upon but was required to keep my sanity while parsing the XML.


wvanbergen:
This will need rebasing. We actually rewrote the UPS carrier using Nokogiri, so don't be afraid of your sanity.

PR #204, opened 2014-12-01
Add USPS handling for "delivered to an individual at the address"

ajlai:
Marks tracking info for links such as https://tools.usps.com/go/TrackConfirmAction?qtc_tLabels1=9405515901127581242940 as delivered correctly.


wvanbergen:
- Can you rebase on the latest master? The USPS carrier was reimplemented using Nokogiri
- Can you add an XML fixture of a response like this?

trishume:
This issue was fixed with the update to the latest USPS tracking API in #260.

PR #210, opened 2015-01-06
Don't load all of activesupport

wvanbergen:
We only use a small subset of active_support. Let's only load what we're actually using, where we are using it. I also removed the autoload for the vendored libraries; we can just require them.

@robfoster @RichardBlair


RichardBlair:
Really like this. :+1:

PR #213, opened 2015-01-07
CONTRIBUTING guidelines

wvanbergen:
@RichardBlair @robfoster as discussed.


wvanbergen:
I'm inclined to use nokogiri for XML, but REXML/Builder is what currently is being used.

RichardBlair:
nokogiri is also what I would like to see, however as long as it's consistent this works. :shipit:

PR #214, opened 2015-01-07
Restrict dependency versions in gemspec

wvanbergen:
This restricts the dependency to certain version ranges.

- Most of the versions I copied from activemerchant. 
- I removed `json` because it is part of the standard library in Ruby 1.9+. We dropped support for Ruby 1.8

@robfoster @RichardBlair


RichardBlair:
:shipit:

PR #215, opened 2015-01-08
Fix Gemfiles to make sure the right versions of activesupport are loaded

wvanbergen:
@kmcphillips @RichardBlair @robfoster FYI.


kmcphillips:
:+1:

PR #218, opened 2015-01-17
Canada Post with nokogiri

wvanbergen:
This updates the legacy Canada Post carrier to use Nokogiri to generate XML requests and parse XML responses, instead of XmlNode and REXML respectively.

I'd like to get rid of XmlNode and standardize on Nokogiri for XML processing. This is one step in that direction. The other option is to standardize on REXML + Builder. Nokogiri is fasterand built on top of a battle tested libxml, but REXML + Builder are pure Ruby. What do you think?

@Shopify/shipping

### DIFF
--- a/lib/active_shipping.rb
+++ b/lib/active_shipping.rb
@@ -27,6 +27,9 @@ require 'active_support/core_ext/enumerable'
 require 'active_support/core_ext/class/attribute_accessors'
 require 'active_utils'
 
+require 'rexml/document'
+require 'nokogiri'
+
 require 'vendor/quantified/lib/quantified'
 require 'vendor/xml_node/lib/xml_node'
 
--- a/test/unit/carriers/canada_post_test.rb
+++ b/test/unit/carriers/canada_post_test.rb
@@ -100,11 +100,13 @@ class CanadaPostTest < Minitest::Test
   end
 
   def test_build_line_items
-    xml_line_items = @carrier.send(:build_line_items, @line_items)
-    assert_instance_of XmlNode, xml_line_items
+    line_items_xml = Nokogiri::XML::Builder.new do |xml|
+      @carrier.send(:build_line_items, xml, @line_items)
+    end
 
-    xml_string = xml_line_items.to_s
-    assert_match /a box full of stuff/, xml_string
+    assert_instance_of Nokogiri::XML::Builder, line_items_xml
+    assert_equal "a box full of stuff", line_items_xml.doc.at_xpath('//description').text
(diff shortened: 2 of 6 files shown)


test/unit/carriers/canada_post_test.rb
```diff
@@ -130,9 +132,4 @@ def test_delivery_range_with_invalid_date
 
     assert_equal [], rate_estimates.rates[0].delivery_range
   end
-
-  def test_line_items_with_nil_values
-    @line_items << Package.new(500, [2, 3, 4], :description => "another box full of stuff", :value => nil)
-    @carrier.find_rates(@origin, @destination, @line_items)
-  end
```
wvanbergen: This tests looks unfinished to me. The HTTP request is not stubbed out (so it executes a slow HTTP request), and we don't do any assertions, so this test will always pass.

I decided to just remove it.

wvanbergen:
Note: the remote tests pass as well.

lib/active_shipping.rb
```diff
@@ -27,6 +27,9 @@
 require 'active_support/core_ext/class/attribute_accessors'
 require 'active_utils'
 
+require 'rexml/document'
+require 'nokogiri'
+
 require 'vendor/quantified/lib/quantified'
 require 'vendor/xml_node/lib/xml_node'
```
garethson: Do we still need this line then?

wvanbergen: Yeah, unfortunately this is not the only one that uses XmlNode.

test/unit/carriers/canada_post_test.rb
garethson: I'm assuming this test used to check if an exception was thrown. I'd rather we keep stub it out the same way we do other similar tests and assert no exception is raised.

garethson:
:+1: for this.

lib/active_shipping.rb
garethson: Oh, duh. Well just convert those too then. :trollface:

wvanbergen: Yup, I need credentials so I can run the remote tests though :)

kmcphillips:
Nokogiri is widely used and massively faster. Lets use it and stick with it.

wvanbergen:
Cool, I am merging this!

test/unit/carriers/canada_post_test.rb
wvanbergen: I re-added this test, stubbed to response XML and added a basic assertion.

PR #220, opened 2015-01-19
Fedex: Convert XML processing to nokogiri

wvanbergen:
Build XML, and parse responses using Nokogiri instead of XmlNode / REXML.

Notes:
- To determine success, we not look at the HighestSeverity tag instead of passing the first notification. This means that the request without notifications now does not raise an exception. The test has been 
- The unit tests pass, but there is on failing remote tests, due to the fact that the XML response for tracking details is very different from the fixture file we use for unit tests. What do do? 

@Shopify/shipping for review please.


wvanbergen:
Ping @jimmybaker because I think you once contributed this :)

test/remote/fedex_test.rb
```diff
@@ -189,7 +189,7 @@ def test_beverly_hills_to_london
   end
 
   def test_tracking
-    p response = @carrier.find_tracking_info('123456789012', :test => true)
```
kmcphillips: :+1:

kmcphillips:
Nothing stands out, but it's a huge diff. Have to trust your tests.

About the remote tests I'm not sure. Any idea where the discrepancy comes from?

wvanbergen:
It looks like the response Fedex sends back for tracking details is different from the response we have as fixtures. So while the unit test with the stubbed response work fine, testing against a real response causes errors.

jimmybaker:
@wvanbergen - Yes I contributed this a long time ago.

Perhaps the tracking response you're getting is using an updated api version. IIRC it was version 3. Is that the same version you get in your remote response?

wvanbergen:
Because the responses I was getting from the v3 tracking API were different from the one in the fixtures, I updated the FedEx tracking API to v7. Documentation is still available for that version.

That means I replaced all fixtures with new fixtures based on test tracking numbers I managed to find on the internet. This means that we now have useful remote tests for this API call as well. However, I may not have been able to replicate all v3 fixtures in v7. I did my best to make sure the response object is compatible with the previous one, to make sure this is backwards compatible though.

@garethson @jimmybaker @kmcphillips @joakimk Thoughts?

lib/active_shipping/carriers/fedex.rb
```diff
@@ -557,22 +562,23 @@ def extract_address(document, possible_node_names)
...
 
     def extract_timestamp(document, node_name)
-      if timestamp_node = document.elements[node_name]
-        Time.parse(timestamp_node.to_s).utc
+      if timestamp_node = document.at(node_name)
+        if timestamp_node.text =~ /\A(\d{4}-\d{2}-\d{2})T00:00:00\Z/
+          Date.parse($1)
+        else
+          Time.parse(timestamp_node.text)
+        end
```
wvanbergen: From API manual: "Time package was shipped/tendered over to FedEx. Time  portion will be populated if available, otherwise will be set to midnight"

wvanbergen: There is no chance of confusing this with a real timestamp that was exactly at midnight, because real timestamps have a timezone included.

RichardBlair:
:tophat:ed on Shopify.

PR #221, opened 2015-01-20
UPS: port XML processing to Nokogiri

wvanbergen:
Except for some tests that are skipped, remote tests pass as well.

@Shopify/shipping

### DIFF
--- a/test/remote/ups_test.rb
+++ b/test/remote/ups_test.rb
@@ -21,7 +21,7 @@ class UPSTest < Minitest::Test
   end
 
   def test_tracking_with_another_number
-    @carrier.find_tracking_info('1Z12345E6692804405')
+    response = @carrier.find_tracking_info('1Z12345E6692804405')
     assert response.success?
   end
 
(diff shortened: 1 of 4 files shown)


test/remote/ups_test.rb
```diff
@@ -21,7 +21,7 @@ def test_tracking_with_bad_number
   end
 
   def test_tracking_with_another_number
-    @carrier.find_tracking_info('1Z12345E6692804405')
+    response = @carrier.find_tracking_info('1Z12345E6692804405')
```
garethson: :+1:

test/remote/ups_test.rb
```diff
@@ -38,7 +38,7 @@ def test_us_to_uk
   end
 
   def test_puerto_rico
-    @carrier.find_rates(
+    response = @carrier.find_rates(
```
garethson: :+1:

garethson:
Diff was a bit tricky to follow but everything looks good. :boat:

kmcphillips:
Nothing stands out. :+1:

wvanbergen:
Now also XML parsing using Nokogiri,

garethson:
@wvanbergen One thing to watch for on these conversions is to ensure that the test data we're parsing has the XML namespaces that production API calls do, otherwise Nokogiri chokes on parsing.

wvanbergen:
Looks like it's fine; remote tests are happy.

wvanbergen:
AFAIK only fedex has namespaces, and we actively remove them before parsing the XML using `gsub` :poop:.

PR #222, opened 2015-01-22
Document ActiveShipping::Carrier using yardoc.

wvanbergen:
This adds Yardoc documentation to the `ActiveShipping::Carrier` class.

Some of the helper classes and response classes definitely need documentation too, but given that the Carrier primarily defines the abstraction, this is a good place to start.

Once we do a release, the docs should show up here: http://www.rubydoc.info/gems/active_shipping

@kmcphillips @jnormore @garethson

### DIFF
--- a/lib/active_shipping/carrier.rb
+++ b/lib/active_shipping/carrier.rb
@@ -1,7 +1,22 @@
 module ActiveShipping
+
+  # Carrier is abstract the base class for all supported carriers.
+  #
+  # To implement support for a carrier, you should subclass this class and
+  # implement all the methods the carrier supports.
+  #
+  # @see #find_rates
+  # @see #create_shipment
+  # @see #find_tracking_info
+  #
+  # @!attribute test_mode
+  #   Whether to interact with the carrier's sandbox environment.
+  #   @return [Boolean]
+  #
+  # @!attribute last_request
+  #   The last request performed against the carrier's API.
+  #   @see #save_request
   class Carrier
-    include ActiveUtils::RequiresParameters
-    include ActiveUtils::PostsData
     include Quantified
 
     attr_reader :last_request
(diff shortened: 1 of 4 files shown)


lib/active_shipping/carrier.rb
```diff
@@ -1,7 +1,22 @@
 module ActiveShipping
+
+  # Carrier is abstract the base class for all supported carriers.
+  #
+  # To implement support for a carrier, you should subclass this class and
+  # implement all the methods the carrier supports.
+  #
+  # @see #find_rates
+  # @see #create_shipment
```
jnormore: create_shipment isn't a mandatory method, some don't support label printing

kmcphillips:
:+1:

lib/active_shipping/carrier.rb
wvanbergen: Yeah, some implement it to return a checking number only. And some don't implement it at all. This is why it says "you should subclass this class and implement all the methods the carrier supports." The `find_tracking_info` method was previously not included in the base class at all, but is implemented by many carriers.

The abstraction is pretty poorly defined (a trait shared with the other abstraction layer gems we have)

wvanbergen:
In general, the abstraction of ActiveShipping is pretty poorly specified. E.g.: parameter and return types, whether a method is required to be implemented or not, etc. This is step one into making that a bit more explicit, so it's easier to use this library and easier to add new carriers.

Already there are inconsistencies between the different carriers. E.g. `create_shipment` sometimes returns a `ShippingResponse`, but sometimes a `LabelResponse`.

lib/active_shipping/carrier.rb
jnormore: :+1:

PR #224, opened 2015-01-23
Convert USPS to nokogiri

wvanbergen:
@Shopify/shipping For review. Remote tests work as well.

### DIFF
--- a/lib/active_shipping/carriers/usps.rb
+++ b/lib/active_shipping/carriers/usps.rb
@@ -168,15 +168,15 @@ module ActiveShipping
       /Delivery status information is not available/
     ]
 
-    ESCAPING_AND_SYMBOLS = /&amp;lt;\S*&amp;gt;/
-    LEADING_USPS = /^USPS/
+    ESCAPING_AND_SYMBOLS = /&lt;\S*&gt;/
+    LEADING_USPS = /^USPS /
     TRAILING_ASTERISKS = /\*+$/
     SERVICE_NAME_SUBSTITUTIONS = /#{ESCAPING_AND_SYMBOLS}|#{LEADING_USPS}|#{TRAILING_ASTERISKS}/
 
     def find_tracking_info(tracking_number, options = {})
       options = @options.update(options)
       tracking_request = build_tracking_request(tracking_number, options)
-      response = commit(:track, tracking_request, (options[:test] || false))
+      response = commit(:track, tracking_request, options[:test] || false)
       parse_tracking_response(response, options)
     end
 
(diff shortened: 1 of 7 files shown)


lib/active_shipping/carriers/usps.rb
```diff
@@ -168,15 +168,15 @@ class USPS < Carrier
       /Delivery status information is not available/
     ]
 
-    ESCAPING_AND_SYMBOLS = /&amp;lt;\S*&amp;gt;/
-    LEADING_USPS = /^USPS/
+    ESCAPING_AND_SYMBOLS = /&lt;\S*&gt;/
+    LEADING_USPS = /^USPS /
```
wvanbergen: These changes were needed due to small parsing differences between Nokogiri and REXML.

lib/active_shipping/carriers/usps.rb
```diff
@@ -168,15 +168,15 @@ class USPS < Carrier
...
     TRAILING_ASTERISKS = /\*+$/
     SERVICE_NAME_SUBSTITUTIONS = /#{ESCAPING_AND_SYMBOLS}|#{LEADING_USPS}|#{TRAILING_ASTERISKS}/
 
     def find_tracking_info(tracking_number, options = {})
       options = @options.update(options)
       tracking_request = build_tracking_request(tracking_number, options)
-      response = commit(:track, tracking_request, (options[:test] || false))
+      response = commit(:track, tracking_request, options[:test] || false)
```
kmcphillips: What is `options[:test] || false` trying to accomplish? Coercing to a true/false class? Because it will only do so for the false case but the true case will pass through. Or to simplify, this means that every value will be passed through without changing, except for `nil` which will be turned into `false`.

kmcphillips: If it's a boolean coercion it should be `!!options[:test]`

lib/active_shipping/carriers/usps.rb
```diff
@@ -432,20 +436,24 @@ def rates_from_response_node(response_node, packages, options = {})
...
+
+        package_node.xpath(service_node).each do |service_response_node|
+          service_name = service_response_node.at(service_name_node).text
 
-        package_node.each_element(service_node) do |service_response_node|
-          service_name = service_response_node.get_text(service_name_node).to_s
+          # service_name.gsub!('&lt;sup&gt;&amp;reg;&lt;/sup&gt;', '') # '®'
+          # service_name.gsub!('&lt;sup&gt;&#8482;&lt;/sup&gt;', '') # '™'
+          # service_name.gsub!('&lt;sup&gt;&amp;trade;&lt;/sup&gt;', '') # '™'
+          # service_name.gsub!(/\*+\z/, '') # '**'
```
kmcphillips: ?

kmcphillips:
Couple of comments. Mostly :+1:

lib/active_shipping/carriers/usps.rb
garethson: :eyes:

garethson:
This LGTM. Are we ever :tophat: this stuff? We probably should in Shopify anyway when we update our gem versions after this ships.

lib/active_shipping/carriers/usps.rb
wvanbergen: I don't like the fact that we have the `:test` option here in the first place. We should be able to pass `:test => true` (or preferably: `environment: :sandbox` or something) to the initializer, and not allow it to be overriden in every method.

lib/active_shipping/carriers/usps.rb
wvanbergen: This was me investigating some slight differences between Nokogiri and REXML. Forgot to remove it.

lib/active_shipping/carriers/usps.rb
wvanbergen: Anyway, that's for a different PR. I changed it to be simply `options[:test]`. Boolean coercion is not necessary here. It's use din `commit`:
``` ruby
    def commit(action, request, test = false)
      ssl_post("#{test ? TEST_URL : LIVE_URL}/#{RESOURCES[action]}", request)
    end
```

wvanbergen:
The changes that got merged so far should be pretty safe. The only one that is more invasive is the FedEx one (#220).

My proposal:
- Merge the "safe" PRs: this one, and #226 and #225 
- Do another prerelease
- Update Shopify & test
- Do the same with Fedex in isolation.

PR #225, opened 2015-01-23
Convert shipwire to nokogiri

wvanbergen:
@Shopify/shipping surprise! Another one. All tests pass.

### DIFF
--- a/test/unit/carriers/shipwire_test.rb
+++ b/test/unit/carriers/shipwire_test.rb
@@ -126,7 +126,7 @@ class ShipwireTest < Minitest::Test
   end
 
   def test_rate_request_includes_address_name_if_provided
-    name = CGI.escape("<Name><Full>Bob Bobsen</Full></Name>")
+    name = CGI.escape("<Full>Bob Bobsen</Full>")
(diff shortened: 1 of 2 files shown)


test/unit/carriers/shipwire_test.rb
```diff
@@ -126,7 +126,7 @@ def test_validate_credentials_with_invalid_credentials
   end
 
   def test_rate_request_includes_address_name_if_provided
-    name = CGI.escape("<Name><Full>Bob Bobsen</Full></Name>")
+    name = CGI.escape("<Full>Bob Bobsen</Full>")
```
garethson: What's up with this?

garethson:
:tada:!

test/unit/carriers/shipwire_test.rb
wvanbergen: The `<Name>` tag is still there but with a bunch of whitespace in between. Given that it just wants to make sure the name is included I though this part would be enough to verify that.

garethson: :+1:

PR #226, opened 2015-01-23
Convert stamps to nokogiri

wvanbergen:
- Generate XML using Nokogiri
- Also, stamps.com actually managed to update their SSL implementation, so we no longer require SSLv3.


RichardBlair:
Seems sane. :shipit:

PR #228, opened 2015-01-26
Rdoc for response classes

wvanbergen:
This adds some documentation for the different response classes.

I tried to do this as well as possible, but the different carriers are not necessarily consistent in what values they assign to the different attributes in the response objects. Not sure if we want to tackle these; for now I left the scope of this PR to docs only.

@Shopify/shipping


kmcphillips:
:+1:

garethson:
:boat:

PR #229, opened 2015-01-26
Replace vendored quantified by the released gem

### LINKED ISSUE #227: Extract Quantified as gem
Right now we have vendored *Quantified* in this library. I porpose to extract this into a separate gem, and add it as an external dependency. The name is still available: https://rubygems.org/gems/quantified

@Shopify/shipping thoughts?

wvanbergen:
@Shopify/shipping This fixes #227.


kmcphillips:
https://rubygems.org/gems/quantified
:rocket:

PR #230, opened 2015-01-26
Fixes a remote tests for USPS: raise an exception if the tracking number is not found.

wvanbergen:
@Shopify/shipping This one fell trough the cracks.

### DIFF
--- a/lib/active_shipping/carriers/usps.rb
+++ b/lib/active_shipping/carriers/usps.rb
@@ -540,7 +540,12 @@ module ActiveShipping
         tracking_details = xml.root.xpath('TrackInfo/TrackDetail')
 
         tracking_summary = xml.root.at('TrackInfo/TrackSummary')
-        tracking_details << tracking_summary
+        if tracking_details.length > 0
+          tracking_details << tracking_summary
+        else
+          success = false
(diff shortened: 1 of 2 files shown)


wvanbergen:
Apparently USPS changed the message it sends when a tracking ID is not found. Instead of testing against a specific error message, I now test against the absence of tracking details instead.

lib/active_shipping/carriers/usps.rb
```diff
@@ -540,7 +540,12 @@ def parse_tracking_response(response, options)
         tracking_details = xml.root.xpath('TrackInfo/TrackDetail')
 
         tracking_summary = xml.root.at('TrackInfo/TrackSummary')
-        tracking_details << tracking_summary
+        if tracking_details.length > 0
+          tracking_details << tracking_summary
+        else
+          success = false
+          meessage = tracking_summary.text
```
kitt: meessage?  not message?

wvanbergen: Fixed!

wvanbergen:
Any other comments before I merge?

kmcphillips:
:+1: Looks less fragile.

PR #231, opened 2015-01-27
Update test framework so remote tests can be run on CI

wvanbergen:
Hello @Shopify/shipping, it's me again! :) 

- Pass credentials.yml through ERB, so we can populate it with values from environment variables.
- Skip remote tests when asking for credentials, but none are set in `credentials.yml`.
- Make sure remote test class names are prefixed with Remote, and do not clash with unit test class names.
- Set up Travis to also run remote tests.

I added the test API keys for Canada Post and New Zealand Post to the Travis environment. These credentials were already made public in `credentials.yml`, so I assume that's OK. For the other ones, we can use [Travis's encrypted environment variables](http://docs.travis-ci.com/user/environment-variables/#Secure-Variables).

### DIFF
--- a/test/remote/canada_post_pws_platform_test.rb
+++ b/test/remote/canada_post_pws_platform_test.rb
@@ -3,7 +3,7 @@
 require 'test_helper'
 
 # All remote tests require Canada Post development environment credentials
-class CanadaPostPWSPlatformTest < Minitest::Test
+class RemoteCanadaPostPWSPlatformTest < Minitest::Test
   include ActiveShipping::Test::Credentials
   include ActiveShipping::Test::Fixtures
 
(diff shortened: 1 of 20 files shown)


test/remote/canada_post_pws_platform_test.rb
```diff
@@ -95,6 +95,8 @@ def setup
     @customer_number  = @login[:customer_number]
     @customer_api_key = @login[:customer_api_key]
     @customer_secret  = @login[:customer_secret]
+  rescue NoCredentialsFound => e
+    skip(e.message)
```
kmcphillips: :+1:

kmcphillips:
Looks gread.

PR #232, opened 2015-01-29
Parse "RescheduledDeliveryDate" from ups tracking response

eboswort:
When an exception delays a UPS shipment, the ScheduledDeliveryDate node is no longer present. Instead there is a "RescheduledDeliveryDate" node, which has the new scheduled delivery date.

That node should be used in the absence of a ScheduledDeliveryDate node.


wvanbergen:
This LGTM. @Shopify/shipping ?

kmcphillips:
:+1: 

Thanks for adding the test and fixture.

eboswort:
Whoops. Failed on Travis. Just pushed what I think will fix that.

PR #233, opened 2015-01-29
Fix Canada Post PWS remote tests

wvanbergen:
- This fixes `CanadaPostPWSTest`
- This does *not* fix `CanadaPostPWSPlatformTest` (yet).
- I committed publicly available test credentials, so we can run the remote tests as part of our CI runs.
- The public credentials are not "platform-enabled", so these cannot be used for the platform test. I also moved two tests from the PWS test to the PWSPlatformTest, because they secretly require platform credentials.

To get the tests to pass, I had to change some assertions here and there. In the implementation itself, I made two changes:

- Don't unnormalize XML. This is not necessary and causes the resulting string to be not parseable in some cases. This was maybe to work around an XML encoding issue in the Canada Post responses that has been fixed since?
- Packages are required to have dimensions. If not, I now mark the shipment as documents.

@Shopify/shipping


kmcphillips:
:+1:

RichardBlair:
10/10 would :+1: again.

PR #235, opened 2015-01-29
Remove builder and XmlNode dependencies

wvanbergen:
- XmlNode is removed from the repository
- Also, This removes the `builder` dependency
- Removes `Location#to_xml`. This requires builder to work. Given that every carrier uses a different way to encode addresses, keeping out "generic" XML representation, and thus the builder dependency, around seems unnecessary to me.


wvanbergen:
@Shopify/shipping

wvanbergen:
Depends on #236.

RichardBlair:
I'll :tophat: by end of week. Bug me on Friday if I don't get to it before then.

PR #236, opened 2015-01-30
[Canada Post AWS] Convert XML processing to Nokogiri

wvanbergen:
Pretty straightforward conversion.

@Shopify/shipping

### DIFF
--- a/lib/active_shipping/carriers/canada_post_pws.rb
+++ b/lib/active_shipping/carriers/canada_post_pws.rb
@@ -158,13 +158,14 @@ module ActiveShipping
     # service discovery
 
     def parse_services_response(response)
-      doc = REXML::Document.new(response)
-      service_nodes = doc.elements['services'].elements.collect('service') { |node| node }
+      doc = Nokogiri.XML(response)
+      doc.remove_namespaces!
+      service_nodes = doc.xpath('services/service')
       service_nodes.inject({}) do |result, node|
-        service_code = node.get_text("service-code").to_s
-        service_name = node.get_text("service-name").to_s
-        service_link = node.elements["link"].attributes['href']
-        service_link_media_type = node.elements["link"].attributes['media-type']
+        service_code = node.at("service-code").text
+        service_name = node.at("service-name").text
+        service_link = node.at("link").attributes['href'].value
+        service_link_media_type = node.at("link").attributes['media-type'].value
         result[service_code] = {
           :name => service_name,
           :link => service_link,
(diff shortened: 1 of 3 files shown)


lib/active_shipping/carriers/canada_post_pws.rb
```diff
@@ -158,13 +158,14 @@ def maximum_weight
...
+      doc.remove_namespaces!
+      service_nodes = doc.xpath('services/service')
       service_nodes.inject({}) do |result, node|
-        service_code = node.get_text("service-code").to_s
-        service_name = node.get_text("service-name").to_s
-        service_link = node.elements["link"].attributes['href']
-        service_link_media_type = node.elements["link"].attributes['media-type']
+        service_code = node.at("service-code").text
+        service_name = node.at("service-name").text
+        service_link = node.at("link").attributes['href'].value
```
kmcphillips: If the `href` is missing will it return `NilClass` or a null object that has a `.value` method?

lib/active_shipping/carriers/canada_post_pws.rb
```diff
@@ -175,36 +176,35 @@ def parse_services_response(response)
...
-            :name => node.get_text("option-name").to_s,
-            :required => node.get_text("mandatory").to_s == "false" ? false : true,
-            :qualifier_required => node.get_text("qualifier-required").to_s == "false" ? false : true
-          }
-          option[:qualifier_max] = node.get_text("qualifier-max").to_s.to_i if node.get_text("qualifier-max")
-          result << option
-          result
-        end
+      doc = Nokogiri.XML(response)
+      doc.remove_namespaces!
```
kmcphillips: lolxml

lib/active_shipping/carriers/canada_post_pws.rb
```diff
@@ -175,36 +176,35 @@ def parse_services_response(response)
...
+      service_code = doc.root.at("service-code").text
+      service_name = doc.root.at("service-name").text
+
+      option_nodes = doc.root.xpath('options/option')
+      options = option_nodes.map do |node|
+        option = {
+          :code => node.at("option-code").text,
+          :name => node.at("option-name").text,
+          :required => node.at("mandatory").text == "false" ? false : true,
+          :qualifier_required => node.at("qualifier-required").text == "false" ? false : true
```
kmcphillips: You don't need the ternary here, `node.at("qualifier-required").text != "false"`. But I guess it's a bit harder to read. Or at least is less explicit.

lib/active_shipping/carriers/canada_post_pws.rb
```diff
@@ -442,102 +449,107 @@ def shipment_customs_node(destination, line_items, options)
...
-          option_nodes.inject({}) do |result, node|
-          result[node.get_text("option-code").to_s] = node.get_text("option-price").to_s.to_f
+      receipt[:priced_options] = if option_nodes.length > 0
+        option_nodes.inject({}) do |result, node|
+          result[node.at("option-code").text] = node.at("option-price").text.to_f
           result
           end
       else
-        []
+        {}
```
kmcphillips: ?

kmcphillips:
Some small questions. Nothing stands out though. Thanks!

lib/active_shipping/carriers/canada_post_pws.rb
wvanbergen: This was a small bug in the original implementation. The `if` clause is returning a hash, but the `else` clause is returning an array. Given that it's empty, that probably never caused any problems, but it irked me.

lib/active_shipping/carriers/canada_post_pws.rb
wvanbergen: Yeah, it will raise an exception in that case. I can fix it so it returns `nil` instead of blowing up, but I find it highly unlikely that the `<link>` tag is there without an `href` attribute.

lib/active_shipping/carriers/canada_post_pws.rb
wvanbergen: Yeah, I was lazy; I didn't want to include the namespace in every xpath query. :)

PR #237, opened 2015-01-30
Add more options to UPS create_shipment

eboswort:
This adds some more options to create_shipment for UPS.
I believe these options are necessary for shipping internationally.
They are definitely necessary for making use of paperless_invoice which allows one to ship commercial packages internationally without printing out a customs invoice.

### DIFF
--- a/lib/active_shipping/carriers/ups.rb
+++ b/lib/active_shipping/carriers/ups.rb
@@ -144,8 +144,8 @@ module ActiveShipping
         xml = parse_ship_confirm(confirm_response)
         success = response_success?(xml)
         message = response_message(xml)
-        digest  = response_digest(xml)
         raise message unless success
+        digest  = response_digest(xml)
 
         # STEP 2: Accept. Use shipment digest in first response to get the actual label.
         accept_request = build_accept_request(digest, options)
@@ -182,7 +182,7 @@ module ActiveShipping
       xml_builder = Nokogiri::XML::Builder.new do |xml|
         xml.AccessRequest do
           xml.AccessLicenseNumber(@options[:key])
-          xml.UserId(@options[:password])
+          xml.UserId(@options[:login])
           xml.Password(@options[:password])
         end
       end
@@ -251,18 +251,29 @@ module ActiveShipping
     # * shipper: who is sending the package and where it should be returned
     #     if it is undeliverable.
     # * ship_from: where the package is picked up.
-    # * service_code: default to '14'
-    # * service_descriptor: default to 'Next Day Air Early AM'
+    # * service_code: default to '03'
     # * saturday_delivery: any truthy value causes this element to exist
     # * optional_processing: 'validate' (blank) or 'nonvalidate' or blank
-    #
-    def build_shipment_request(origin, destination, packages, options = {})
-      # There are a lot of unimplemented elements, documenting all of them
-      # wouldprobably be unhelpful.
+    # * paperless_invoice: set to truthy if using paperless invoice to ship internationally
+    # * terms_of_shipment: used with paperless invoice to specify who pays duties and taxes
+    # * reference_numbers: Array of hashes with :value => a reference number value and optionally :code => reference number type
+    # * prepay: if truthy the shipper will be bill immediatly. Otherwise the shipper is billed when the label is used.
+    def build_shipment_request(origin, destination, packages, options={})
+      packages = Array(packages)
+      options[:international] = origin.country.name != destination.country.name
+      options[:imperial] ||= IMPERIAL_COUNTRIES.include?(origin.country_code(:alpha2))
+      if allow_package_level_reference_numbers(origin, destination)
+        if options[:reference_numbers]
+          packages.each do |package|
+            package.options[:reference_numbers] = options[:reference_numbers]
+          end
+        end
+        options[:reference_numbers] = []
+      end
+
       xml_builder = Nokogiri::XML::Builder.new do |xml|
         xml.ShipmentConfirmRequest do
           xml.Request do
-            # Required element and the text must be "ShipConfirm"
             xml.RequestAction('ShipConfirm')
             # Required element cotnrols level of address validation.
             xml.RequestOption(options[:optional_processing] || 'validate')
@@ -275,54 +286,71 @@ module ActiveShipping
           end
 
           xml.Shipment do
-            # Required element.
             xml.Service do
-              xml.Code(options[:service_code] || '14')
-              xml.Description(options[:service_description] || 'Next Day Air Early AM')
+              xml.Code(options[:service_code] || '03')
             end
 
-            # Required element. The delivery destination.
-            build_location_node(xml, 'ShipTo', destination, {})
+            build_location_node(xml, 'ShipTo', destination, options)
+            build_location_node(xml, 'ShipFrom', origin, options)
             # Required element. The company whose account is responsible for the label(s).
-            build_location_node(xml, 'Shipper', options[:shipper] || origin, {})
-            # Required if pickup is different different from shipper's address.
-            build_location_node(xml, 'ShipFrom', options[:ship_from], {}) if options[:ship_from]
+            build_location_node(xml, 'Shipper', options[:shipper] || origin, options)
 
-            # Optional.
             if options[:saturday_delivery]
               xml.ShipmentServiceOptions do
                 xml.SaturdayDelivery
               end
             end
 
-            # Optional.
             if options[:origin_account]
               xml.RateInformation do
                 xml.NegotiatedRatesIndicator
               end
             end
 
-            # Optional.
-            if options[:shipment] && options[:shipment][:reference_number]
+            Array(options[:reference_numbers]).each do |reference_num_info|
               xml.ReferenceNumber do
-                xml.Code(options[:shipment][:reference_number][:code] || "")
-                xml.Value(options[:shipment][:reference_number][:value])
+                xml.Code(reference_num_info[:code] || "")
+                xml.Value(reference_num_info[:value])
               end
             end
 
-            # Conditionally required.  Either this element or an ItemizedPaymentInformation
-            # is needed.  However, only PaymentInformation is not implemented.
-            xml.PaymentInformation do
-              xml.Prepaid do
-                xml.BillShipper do
-                  xml.AccountNumber(options[:origin_account])
+            if options[:prepay]
+              xml.PaymentInformation do
+                xml.Prepaid do
+                  xml.BillShipper do
+                    xml.AccountNumber(options[:origin_account])
+                  end
+                end
+              end
+            else
+              xml.ItemizedPaymentInformation do
+                xml.ShipmentCharge do
+                  # Type '01' means 'Transportation'
+                  # This node specifies who will be billed for transportation.
+                  xml.Type('01')
+                  xml.BillShipper do
+                    xml.AccountNumber(options[:origin_account])
+                  end
+                end
+                if options[:terms_of_shipment] == 'DDP'
+                  # DDP stands for delivery duty paid and means the shipper will cover duties and taxes
+                  # Otherwise UPS will charge the receiver
+                  xml.ShipmentCharge do
+                    xml.Type('02') # Type '02' means 'Duties and Taxes'
+                    xml.BillShipper do
+                      xml.AccountNumber(options[:origin_account])
+                    end
+                  end
                 end
               end
             end
 
+            if options[:international]
+              build_international_shipment_request_options(xml, origin, destination, packages, options)
+            end
+
             # A request may specify multiple packages.
-            options[:imperial] ||= IMPERIAL_COUNTRIES.include?(origin.country_code(:alpha2))
-            Array(packages).each do |package|
+            packages.each do |package|
               build_package_node(xml, package, options)
             end
           end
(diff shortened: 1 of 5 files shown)


lib/active_shipping/carriers/ups.rb
```diff
@@ -182,7 +182,7 @@ def build_access_request
       xml_builder = Nokogiri::XML::Builder.new do |xml|
         xml.AccessRequest do
           xml.AccessLicenseNumber(@options[:key])
-          xml.UserId(@options[:password])
+          xml.UserId(@options[:login])
```
wvanbergen: whoops, nice catch!

eboswort: ha ha yea. None of the remote tests were running!

lib/active_shipping/carriers/ups.rb
```diff
@@ -275,53 +277,75 @@ def build_shipment_request(origin, destination, packages, options = {})
...
             end
 
-            # Optional.
-            if options[:shipment] && options[:shipment][:reference_number]
-              xml.ReferenceNumber do
-                xml.Code(options[:shipment][:reference_number][:code] || "")
-                xml.Value(options[:shipment][:reference_number][:value])
+            if options[:reference_numbers]
+              # UPS only allows shipment-level reference numbers if the package is not US -> US or PR -> PR
+              unless [['US','US'],['PR', 'PR']].include?([origin,destination].map(&:country_code))
```
wvanbergen: can you move this logic to a helper method with a descriptive name?

wvanbergen: Also, unsure about this, but should we raise an exception (ArgumentError) if somebody specifies reference numbers, but UPS doesn't support it?

lib/active_shipping/carriers/ups.rb
```diff
@@ -387,7 +457,7 @@ def build_location_node(xml, name, location, options = {})
           xml.ShipperAssignedIdentificationNumber(destination_account)
         end
 
-        if name = location.company_name || location.name
+        if name = location.company_name || location.name || options[:origin_name]
```
wvanbergen: Can you add parentheses for consistency here?

test/credentials.yml
```diff
@@ -11,9 +11,11 @@ usps:
...
 
 ups:
-  key: <%= ENV['ACTIVESHIPPING_UPS_LOGIN'] %>
-  login: <%= ENV['ACTIVESHIPPING_UPS_KEY'] %>
-  password: <%= ENV['ACTIVESHIPPING_UPS_PASSWORD'] %>
+  key: 5CD7DE7593CBF516
+  login: apicasperuser
+  password: T30roFStkUyn
+  origin_account: W5A760
+  origin_name: Casper
```
wvanbergen: Is this intentional? If so, can we use these to run the remote tests in CI? In this case, we should add them to `.travis.yml` instead as environment variables.

I'd prefer to create a separate credentials entry called `ups_origin`, so we can see what tests/functionality requires an origin account, and which one are not.

test/remote/ups_test.rb
```diff
@@ -256,4 +255,28 @@ def test_obtain_shipping_label_without_dimensions
     # ought to raise an exception and this test will fail.
     assert_instance_of ActiveShipping::LabelResponse, response
   end
+
+  def test_obtain_international_shipping_label
+    skip '<#<RuntimeError: Could not obtain shipping label. Invalid Access License number.>>.'
```
wvanbergen: Why is this skipped?

wvanbergen:
Some comments and questions, looks like a great addition.

test/remote/ups_test.rb
eboswort: This line is just me copying what is done in the previous test. I believe the intention is to skip remote tests for which there are not local credentials. Although this is also the reason the typo of passing in the password as the user id that I fixed in this pr didn't cause failed tests but rather just caused those tests to get skipped. I'm happy to remove it but it would cause tests to fail for people that don't have access to the ups api.

test/credentials.yml
eboswort: This was me being stupid. I went ahead and changed my password and deactivated that key.

eboswort:
Thanks for the comments. I'll go ahead and do what you suggest at some point this weekend or on Monday morning and push an update.

eboswort:
Ok I cleaned up the reference number thing. You have to put the reference numbers in a different place in the xml based on the origin and destination. I tried to make it more explicit what's happening. If you pass in options[:reference_numbers] it will stick them in the right place by default based on the origin and destination.

lib/active_shipping/carriers/ups.rb
```diff
@@ -251,18 +251,29 @@ def build_rate_request(origin, destination, packages, options = {})
     # * shipper: who is sending the package and where it should be returned
     #     if it is undeliverable.
     # * ship_from: where the package is picked up.
-    # * service_code: default to '14'
-    # * service_descriptor: default to 'Next Day Air Early AM'
+    # * service_code: default to '03'
```
wvanbergen: Why this change in default?

lib/active_shipping/carriers/ups.rb
wvanbergen: Interestingly they work fine for me on the master branch :/

wvanbergen:
I'd like tor run the remote tests for the stuff that requires an origin account before merging. Unfortunately I do not have test credentials I can use. Any chance you can share them with me in a secure way? Ping me at willem AT shopify DOT com

lib/active_shipping/carriers/ups.rb
eboswort: UPS Ground seemed a far more reasonable default to me than Next Day Early AM, which is wildly expensive and probably only used in emergencies. Usually UPS Ground would be the default shipping option if you were ordering from a website that was shipping via ups and I imagine would be the default in most other situations as well. I'm assuming the reason it was 14 is because that's what is used in the example in UPS's documentation.

wvanbergen: Sounds good :+1:

wvanbergen: Sounds good :+1:

test/remote/ups_test.rb
wvanbergen: Sounds good :+1:

eboswort:
Anything else I can do to get this merged?

eboswort:
@Shopify/shipping So.... Can we merge? Can we merge? Or like what else do you want me to do?

robfoster:
Greetings @eboswort,

Thanks so much for contributing to the project. Yes let's definitely get this merged. My only reservation is the disabled tests. Somehow we need to get creds that we can share so that we can run these tests.

If we can't get the creds in the next day or two, let's merge this anyways so you're not blocked. I just want to make sure we don't introduce untestable code into the system. 

Thoughts?

eboswort:
Hi Rob! 
Thanks for having me as a contributor. :)
I totally understand you not wanting a bunch of skipped tests on master. But I am definitely trying to start using this stuff pretty soon. If you guys want more time then let me know and I can just work off a fork in the meantime. 
What problems are you running into getting credentials?
UPS has pretty awful support for testing. I just confirmed with them that there's no way to create a testing shipper account. But I think I know the necessary steps to create an empty account for active_shipping that could be used only by travis CI for testing. 
You can create a free account on ups.com and also create a shipper account. As long as you don't select an automatic pickup option for that latter then I'm pretty sure that's free too. So if you do those two things and don't give ups any billing info then you could just share those credentials publicly. I would do it, but I figure it should probably be connected to a shopify email address since you guys are managing this thing.
Or did you try something along those lines and run into trouble?

robfoster:
I created a new UPS account, which I'm going to share with you. Still getting test failures. Are they passing with your account?

### Test results
```
07:26:55 {CasperSleep-international_ups_shipping} ~/src/active_shipping$  bundle exec ruby -Ilib:test test/remote/ups_test.rb 
Run options: --seed 15186

# Running:

..F...S......E.FS

Finished in 9.473338s, 1.7945 runs/s, 3.9057 assertions/s.

  1) Failure:
RemoteUPSTest#test_obtain_shipping_label [test/remote/ups_test.rb:220]:
test/fixtures.yml must have a valid ups/origin_account for this test to run


  2) Error:
RemoteUPSTest#test_obtain_shipping_label_without_dimensions:
RuntimeError: Could not obtain shipping label. Failure.
    /home/vagrant/src/active_shipping/lib/active_shipping/carriers/ups.rb:162:in `rescue in create_shipment'
    /home/vagrant/src/active_shipping/lib/active_shipping/carriers/ups.rb:132:in `create_shipment'
    test/remote/ups_test.rb:243:in `test_obtain_shipping_label_without_dimensions'


  3) Failure:
RemoteUPSTest#test_obtain_international_shipping_label [test/remote/ups_test.rb:262]:
test/fixtures.yml must have a valid ups/origin_account for this test to run

17 runs, 37 assertions, 2 failures, 1 errors, 2 skips

You have skipped tests. Run with --verbose for details.
```

### Creds
```
 cat ~/.active_shipping/credentials.yml 
ups:
  key: "7CE85DED4C9D07AB"
  login: "shopifolk"
  password: "Shopify_rocks"
```

eboswort:
Looks like that's failing cause you don't have a shipper account (sometimes referred to as an origin account). A shipper account is a separate type of account that UPS has. It can be shared and used by multiple ups.com accounts, which when given access, can bill that account for shipments that they create.
It is also sometimes referred to as the origin account, which is what those messages are talking about when they say "must have a valid ups/origin_account".
You can create one on ups.com. You just login, then click the shipping tab up top. Then click "Open a Shipping Account". Make sure to select that you don't want automatic pickup, because UPS charges a weekly fee for that.
If you create one they will give you a 6-digit account number that is your shipper account number.
Then just add a line to credentials.yml so it looks like:
ups:
  key: "7CE85DED4C9D07AB"
  login: "shopifolk"
  password: "Shopify_rocks"
  origin_account: <Your Shipper Account Number>

Those 3 should pass if you do that.

robfoster:
Down to two errors.

```
  1) Error:
RemoteUPSTest#test_obtain_shipping_label:
RuntimeError: Could not obtain shipping label. Failure.
    /home/vagrant/src/active_shipping/lib/active_shipping/carriers/ups.rb:162:in `rescue in create_shipment'
    /home/vagrant/src/active_shipping/lib/active_shipping/carriers/ups.rb:132:in `create_shipment'
    test/remote/ups_test.rb:223:in `test_obtain_shipping_label'


  2) Error:
RemoteUPSTest#test_obtain_shipping_label_without_dimensions:
RuntimeError: Could not obtain shipping label. Failure.
    /home/vagrant/src/active_shipping/lib/active_shipping/carriers/ups.rb:162:in `rescue in create_shipment'
    /home/vagrant/src/active_shipping/lib/active_shipping/carriers/ups.rb:132:in `create_shipment'
    test/remote/ups_test.rb:243:in `test_obtain_shipping_label_without_dimensions'

ups:
  key: "7CE85DED4C9D07AB"
  login: "shopifolk"
  password: "Shopify_rocks"
  origin_account: "X3V606"
```

robfoster:
@RichardBlair, any thoughts?

eboswort:
Awesome, we're getting there!

Sorry but I've had another credential locally that is required which is 
origin_name: Shopify

I think this "origin_name" was required for creating shipments prior to my PR but maybe not. Obviously it's not great that it's so difficult to figure out what options are required to actually get label creation to work. Part of that is the fault of response_message which sometimes (maybe always) fails to pull out the informative part of the error message, which in this case would be helpful in figuring out what's wrong.
Any suggestions as to what to do about origin_name and origin_account not being obvious requirements? I could just throw a big fat error if they are not there when attempting to create labels.

After adding this credential I found these tests were still failing. This was because negotiated rates indicator was automatically turned on if you passed in an origin_account. I did not account for the situation where someone has a shipper account but does not have negotiated rates with UPS. To remedy that I just pushed another commit that makes negotiated rates optional.

If you attempt the tests with the origin_name credential and my latest commit you should see both those tests pass. However now I'm seeing "test_just_country_given_with_origin_account_fails" and "test_ottawa_to_us_fails_with_only_zip_and_origin_account" fail. I suppose some extra information about the shipper that is now being provided is causing those tests to successfully retrieve rates. 
I don't wanna be the guy who wants to remove tests rather than fix them, but I'l just throw that out as an option. Because why is it important that active_shipping is unable to retrieve rates in those situations? As things are with my latest commit, we are passing that info to ups and they are returning rates. Seems like an OK situation to me. 
Any advice on how to move forward with those failing tests?

robfoster:
Agreed. It's not obvious why we were asserting that UPS shouldn't give rates. Instead of deleting the tests, we could assert that things pass:

```
07:40:13 {CasperSleep-international_ups_shipping} ~/src/active_shipping$ git diff
diff --git a/test/remote/ups_test.rb b/test/remote/ups_test.rb
index 8513ba8..88a5343 100644
--- a/test/remote/ups_test.rb
+++ b/test/remote/ups_test.rb
@@ -63,15 +63,12 @@ class RemoteUPSTest < Minitest::Test
     refute response.rates.empty?
   end
 
-  def test_just_country_given_with_origin_account_fails
-    skip unless @options[:origin_account]
-    assert_raises(ResponseError) do
-      @carrier.find_rates(
-        location_fixtures[:beverly_hills],
-        Location.new(:country => 'CA'),
-        Package.new(100, [5, 10, 20])
-      )
-    end
+  def test_just_country_given_with_origin_account_passes
+    @carrier.find_rates(
+      location_fixtures[:beverly_hills],
+      Location.new(:country => 'CA'),
+      Package.new(100, [5, 10, 20])
+    )
   end
 
   def test_ottawa_to_beverly_hills
@@ -109,17 +106,13 @@ class RemoteUPSTest < Minitest::Test
     assert_nil package_rate[:rate]
   end
 
-  def test_ottawa_to_us_fails_with_only_zip_and_origin_account
-    skip unless @options[:origin_account]
-
-    assert_raises ResponseError do
-      @carrier.find_rates(
-        location_fixtures[:ottawa],
-        Location.new(:country => 'US', :zip => 90210),
-        package_fixtures.values_at(:book, :wii),
-        :test => true
-      )
-    end
+  def test_ottawa_to_us_passes_with_only_zip_and_origin_account
+    @carrier.find_rates(
+      location_fixtures[:ottawa],
+      Location.new(:country => 'US', :zip => 90210),
+      package_fixtures.values_at(:book, :wii),
+      :test => true
+    )
   end
```

This works with and without `origin_name: "Shopify"`

eboswort:
@robfoster re: those suggested changes to the tests. 
Take a look at "test_ottawa_to_us_succeeds_with_only_zip" and "test_just_country_given". 
Those ones already exist and are basically the same as what you've modified the two failing tests to be. 
It seems whoever wrote this, asserted that each of these things work without an origin account and fail with an origin account. I think what we've discovered is that they work when you have no origin_account and also when you have both an origin account and a origin_name.
I'd say let's just remove the skip if @options[:origin_account] from the two tests I just mentioned, delete the two that are failing, and consider origin_account and origin_name a pair that must both be present or absent. And maybe document that somehow or raise an error if you try to specify only one of them?

robfoster:
Agreed: Delete the two tests that are failing because they're asserting failure that doesn't happen when we have the correct input. And document that for UPS tests we need five settings configured. The values below can be published (it's an account we just created for the purpose of getting these tests passing).
```
ups:
  key: "7CE85DED4C9D07AB"
  login: "shopifolk"
  password: "Shopify_rocks"
  origin_account: "X3V606"
  origin_name: "Shopify"```

Are we on the same page?

eboswort:
Yep!
I just removed those tests and added origin_name and origin_account to credentials.yml to signal that those things need to be supplied for ups.
Is there somewhere else I should document those options as required?
Also where were you thinking of publishing that account info? Is that supposed to be part of my PR? Or can you input that info into the travis CI environment?

robfoster:
I'll make sure it gets into Travis. @wvanbergen do you have creds for this?

robfoster:
@eboswort Thanks so much for working with us!

eboswort:
Awesome! Thanks for your help getting it merged @robfoster @wvanbergen

wvanbergen:
I'll set up the credentials on Travis tomorrow.

wvanbergen:
See #244.

@eboswort Thanks again for being patient with us here. It took a while before this got merged, but definitely the state of this project with regards to UPS is much better now. That should help getting your other PRs merged faster ;)

PR #238, opened 2015-02-02
fix null pointer exception in usps validate

robfoster:
### Problem
* Refactoring caused a NPE by calling .text on a non-existant XML path

### Solution
* Add test coverage to the validation method
* Fix the NPE

@Shopify/shipping @wvanbergen

### DIFF
--- a/lib/active_shipping/carriers/usps.rb
+++ b/lib/active_shipping/carriers/usps.rb
@@ -286,7 +286,7 @@ module ActiveShipping
       </CarrierPickupAvailabilityRequest>
       EOF
       xml = Nokogiri.XML(commit(:test, request, true)) { |config| config.strict }
-      xml.at('/CarrierPickupAvailabilityResponse/City').text == 'SAN FRANCISCO' && xml.at('/CarrierPickupAvailabilityResponse/Address2').text == '18 FAIR AVE'
+      xml.at('/CarrierPickupAvailabilityResponse/City').try(:text) == 'SAN FRANCISCO' && xml.at('/CarrierPickupAvailabilityResponse/Address2').try(:text) == '18 FAIR AVE'
     end
 
     # options[:service] --    One of [:first_class, :priority, :express, :bpm, :parcel,
--- a/test/remote/usps_test.rb
+++ b/test/remote/usps_test.rb
@@ -201,8 +201,20 @@ class RemoteUSPSTest < Minitest::Test
     end
   end
 
-  def test_valid_credentials
+  def test_correct_login_passes_valid_credentials?
     assert USPS.new(@usps_credentials.merge(:test => true)).valid_credentials?
+  end
+
+  def test_wrong_login_fails_in_valid_credentials?
+    refute USPS.new(:login => 'ABCDEFGHIJKL', :test => true).valid_credentials?
+  end
+
+  def test_blank_login_fails_in_valid_credentials?
+    refute USPS.new(:login => '', :test => true).valid_credentials?
+  end
+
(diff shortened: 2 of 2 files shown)


lib/active_shipping/carriers/usps.rb
```diff
@@ -286,7 +286,9 @@ def canned_address_verification_works?
       </CarrierPickupAvailabilityRequest>
       EOF
       xml = Nokogiri.XML(commit(:test, request, true)) { |config| config.strict }
-      xml.at('/CarrierPickupAvailabilityResponse/City').text == 'SAN FRANCISCO' && xml.at('/CarrierPickupAvailabilityResponse/Address2').text == '18 FAIR AVE'
+      city = xml.at('/CarrierPickupAvailabilityResponse/City')
+      address = xml.at('/CarrierPickupAvailabilityResponse/Address2')
+      city && city.text == 'SAN FRANCISCO' && address && address.text == '18 FAIR AVE'
```
wvanbergen: FYI: you could also use `city.try(:text) == '...' && address.try(:text) == '...'`. Some people don't like `try`, so feel free to leave it like this.

test/unit/carriers/usps_test.rb
```diff
@@ -451,6 +451,30 @@ def test_extract_event_details_handles_double_digit_calendar_dates
     assert_equal 12, details.zoneless_time.mday
   end
 
+  def test_correct_login_succeeds_in_valid_credentials?
+    @carrier = USPS.new(:login => '677JADED7283')
```
wvanbergen: Can we safely commit this key to a public repo?

test/unit/carriers/usps_test.rb
```diff
@@ -451,6 +451,30 @@ def test_extract_event_details_handles_double_digit_calendar_dates
...
+
+  def test_blank_login_fails_in_valid_credentials?
+    @carrier = USPS.new(:login => '')
+    @carrier.expects(:test_mode?).returns(true)
+    refute @carrier.valid_credentials?
+  end
+
+  def test_nil_login_fails_in_valid_credentials?
+    @carrier = USPS.new(:login => nil)
+    @carrier.expects(:test_mode?).returns(true)
```
wvanbergen: Instead of the `expects` call, just use: `@carrier = USPS.new(:login => nil, :test => true)`

test/unit/carriers/usps_test.rb
```diff
@@ -451,6 +451,30 @@ def test_extract_event_details_handles_double_digit_calendar_dates
...
+  def test_wrong_login_fails_in_valid_credentials?
+    @carrier = USPS.new(:login => 'ABCDEFGHIJKL')
+    @carrier.expects(:test_mode?).returns(true)
+    refute @carrier.valid_credentials?
+  end
+
+  def test_blank_login_fails_in_valid_credentials?
+    @carrier = USPS.new(:login => '')
+    @carrier.expects(:test_mode?).returns(true)
+    refute @carrier.valid_credentials?
```
wvanbergen: Given that you only use the variable locally, no need to use an instance variable. SO you can ditch the `@` in front of `@carrier`.

wvanbergen:
Nice work, just some questions.

wvanbergen:
So, we just need to move the tests to the USPS remote test and this is good to go.

kmcphillips:
:+1:

robfoster:
Thanks for your comments, @wvanbergen. They are now addressed.

wvanbergen:
:shipit:

PR #240, opened 2015-02-11
include error details when a UPS rate request fails

robfoster:
### Problem
We return "failure" when a UPS rate request fails instead of the cause

### Solution
Return the error type and details with failed UPS rate requests


lib/active_shipping/carriers/ups.rb
```diff
@@ -626,7 +626,7 @@ def response_success?(document)
     end
 
     def response_message(document)
-      document.root.at_xpath('Response/Error/ErrorDescription | Response/ResponseStatusDescription').text
+      document.root.at_xpath('Response/ResponseStatusDescription').text << ": " << document.root.at_xpath('Response/Error/ErrorDescription').text
```
RichardBlair: This reminds me how much I hate xpath.

PR #241, opened 2015-02-12
Return response status and error desc on failed rate requests

robfoster:
### Problem
We return "failure" when a UPS rate request fails instead of the cause

### Solution
Return the error type and details with failed UPS rate requests

@Shopify/shipping @wvanbergen


lib/active_shipping/carriers/ups.rb
```diff
@@ -626,7 +626,9 @@ def response_success?(document)
     end
 
     def response_message(document)
-      document.root.at_xpath('Response/Error/ErrorDescription | Response/ResponseStatusDescription').text
+      status = document.root.at_xpath('Response/ResponseStatusDescription').try(:text)
+      desc = document.root.at_xpath('Response/Error/ErrorDescription').try(:text)
+      "#{status}: #{desc}"
```
robfoster: [status, desc].select(&:present?).join(": ")

test/unit/carriers/ups_test.rb
```diff
@@ -140,6 +140,21 @@ def test_find_tracking_info_should_return_correct_rescheduled_delivery_date
     assert_equal Time.parse('2015-01-29 00:00:00 UTC'), response.scheduled_delivery_date
   end
 
+  def test_response_parsing_an_oversize_package
+    mock_response = xml_fixture('ups/package_exceeds_maximum_length')
+    @carrier.expects(:commit).returns(mock_response)
+
+    begin
+      @carrier.find_rates(location_fixtures[:beverly_hills],
```
robfoster: assert_raises ActiveShipping::ResponseError do |ex|???
end

lib/active_shipping/carriers/ups.rb
```diff
@@ -626,7 +626,9 @@ def response_success?(document)
     end
 
     def response_message(document)
-      document.root.at_xpath('Response/Error/ErrorDescription | Response/ResponseStatusDescription').text
+      status = document.root.at_xpath('Response/ResponseStatusDescription').try(:text)
+      desc = document.root.at_xpath('Response/Error/ErrorDescription').try(:text)
+      [status, desc].select(&:present?).join(": ").presence || "UPS could not process the request."
```
garethson: Are there ever actually cases where we just get one of these? Status or description?

lib/active_shipping/carriers/ups.rb
```diff
@@ -626,7 +626,9 @@ def response_success?(document)
     end
 
     def response_message(document)
-      document.root.at_xpath('Response/Error/ErrorDescription | Response/ResponseStatusDescription').text
```
garethson: Your code is an improvement for sure, but does it now actually return the proper description when it wasn't before?

garethson:
:boat: other than my questions.

eboswort:
How are we feeling about this one? I'm sick of these generic Failure messages!
Sorry to be all of a sudden coming after you guys to merge a million things ;)

wvanbergen:
:+1:. Might be a good idea to generate a whole bunch of error responses to get more confidence in that we're generating useful messages for all of them.

lib/active_shipping/carriers/ups.rb
robfoster: Before it would just say "Failure" without returning the description. Now it'll return something like: 
"Failure: Package exceeds the maximum length constraint of 108 inches. Length is the longest side of a package."

lib/active_shipping/carriers/ups.rb
robfoster: Unknown but I want to keep current behavior

robfoster:
Let's improve the error reporting even more in another PR. This fix adds descriptions to generic failure messages.

Issue #242, opened 2015-02-18
Support for UPS Time in Transit / Accurate Estimated Delivery Dates


eboswort:
Hi,

I brought this up a few months back on the google group but nobody ever responded to me.

I've got support for the UPS time in transit API working on a fork of ActiveShipping and I'd love to merge that functionality back into shopify/active_shipping.

I'd like to get some feedback on how I could code this up in a way that other users of ActiveShipping would be down with because it's not totally straightforward. For one thing I'm not sure how this functionality is reflected in other carriers.

In the world of the UPS API, the Time in Transit API is the only way to get an accurate idea of when a shipment will arrive based on the day you send it, the origin and the destination. 

The current implementation of get_rates for UPS will give you a delivery_range and a delivery_date for UPS, but these are pretty inaccurate. They are based on the GuaranteedDaysToDelivery field that is returned by UPS's rating API. But that field doesn't take origin, destination, or pickup date into account. So for UPS ground it will always return 5 because that is the max number of days that UPS ground takes for any origin/destination. This ignores the fact that for any particular origin zipcode, most of the continental United States can be reached in less than 5 days (http://www.ups.com/maps). 
The current calculation of delivery_range/delivery_date also ignores holidays and can sometimes underestimate the delivery time for packages that will actually take more than 5 weekdays.

The time in transit API solves these problems. You tell it your origin, destination and pickup date, it tells you the expected delivery date for each service level. It takes holidays into account. And can help you decide to use a cheaper option like UPS ground, when that option will end up getting the shipment there just as fast as a more expensive one.

So like I said I would like to write up a PR adding support for the UPS time in transit API to shopify/active_shipping.

I could see this working in two ways. 

There could be a new method on carrier called get_delivery_dates (or some better name). This is how I have this functionality implemented in my fork. It returns a DeliveryDatesResponse object which has an array of delivery_date_estimates, one for each service. It's a similar set up to the get_rates stuff. This is working fine for me. I guess the only potential problem is if this functionality doesn't make any sense for any other carriers then maybe it's not so great to have on the carrier class? I could implement it directly on the UPS class. Is there a general active_shipping policy for carrier specific functionality?

Another option would be to implement support for an option passed into get_rates that was like accurate_delivery_estimates => true, which for UPS would cause it to calculate the rates the same way it does now, but then also hit the Time in Transit API and fill in the delivery_ranges of those RateEstimates with more accurate estimated delivery dates.

[...]

PR #243, opened 2015-02-20
Add support for UPS's TimeInTransit API

eboswort:
This PR is one of the options I brought up in #242 to add support for UPS's time in transit API.
 
It adds support for accurate time in transit estimates for UPS shipments.


lib/active_shipping/carriers/ups.rb
```diff
@@ -164,6 +169,16 @@ def create_shipment(origin, destination, packages, options = {})
       end
     end
 
+    def get_delivery_dates(origin, destination, packages, pickup_date=Date.current, options = {})
```
wvanbergen: Can you move pickup date into `options`? This way the method call is compatible with `find_rates`, which would allow us to merge them at some point. Given that it's optional it makes more sense to have in the options hash anyway.

lib/active_shipping/delivery_date_estimate.rb
```diff
@@ -0,0 +1,19 @@
+module ActiveShipping
+  class DeliveryDateEstimate
```
wvanbergen: Does this API call also include rates?

I'd prefer it if this class inherits from `RateResponse`, or at least make it duck type compatible with RateResponse for the same reason I noted above.

lib/active_shipping/carriers/ups.rb
```diff
@@ -684,6 +735,28 @@ def parse_tracking_response(response, options = {})
...
 
+    def parse_delivery_dates_response(origin, destination, packages, response, options={})
+      xml     = build_document(response, 'TimeInTransitResponse')
+      success = response_success?(xml)
+      message = response_message(xml)
+      delivery_estimates = []
+
+      if success
+        xml.css('ServiceSummary').each do |service_summary|
+          # vvvvv Translate the Time in Transit Codes to the service codes used elsewhere vvvvv
```
garethson: `vvvvv` probably a typo?

lib/active_shipping/delivery_date_estimate.rb
garethson: AFAIK It won't include pricing info so I'm ok with this new class.

eboswort: Yea it does not have pricing info.

lib/active_shipping/carriers/ups.rb
eboswort: haha I actually just put that in to draw attention to that comment since I was pissed off that UPS decided to use two different sets of service codes in the API. But I'll take that out.

lib/active_shipping/carriers/ups.rb
eboswort: I'm thinking this no longer applies given our discussion on #242

eboswort:
Ok I renamed get_delivery_dates to get_delivery_date_estimates since that seemed a bit more clear. Also I removed the "vvvvvv"'s from that comment.

Any other comments?

wvanbergen:
LGTM.

PR #245, opened 2015-02-23
Support to Brazilian shipping service Correios

albiere:
Hi Guys,

Well, I added support to a new carrier from Brazil called Correios. So far, I've only made the code to calculate the shipping rates. I'll have a look how to integrate the tracking number feature into it later.

I want to point out that the Correios web service does not require any kind of authentication (yeah, seems bad), so I didn't need to create any env variable to test it (remote tests).

That's all!


test/remote/correios_test.rb
```diff
@@ -0,0 +1,81 @@
...
+    @invalid_city = Location.new(:zip => "invalid")
+
+    @book = Package.new(250, [14, 19, 2])
+    @poster = Package.new(100, [93, 15], :cylinder => true)
+
+    @response_clothes = xml_fixture('correios/clothes_response')
+    @response_shoes = xml_fixture('correios/shoes_response')
+    @response_book_success = xml_fixture('correios/book_response')
+    @response_poster_success = xml_fixture('correios/poster_response')
+    @response_book_invalid = xml_fixture('correios/book_response_invalid')
```
wvanbergen: I don't think these fixtures are used here.

lib/active_shipping/carriers/correios.rb
```diff
@@ -0,0 +1,271 @@
...
+        "nVlComprimento=#{params[:length]}&" +
+        "nVlAltura=#{params[:height]}&" +
+        "nVlLargura=#{params[:width]}&" +
+        "nVlDiametro=#{params[:diameter]}&" +
+        "sCdMaoPropria=#{params[:mao_propria_extra]}&" +
+        "nVlValorDeclarado=#{params[:declared_value_extra]}&" +
+        "sCdAvisoRecebimento=#{params[:delivery_notice_extra]}&" +
+        "nIndicaCalculo=#{params[:return_information]}&" +
+        "StrRetorno=#{params[:return_type]}"
+      end
```
wvanbergen: This approach could potentially lead to an invalid URL, based on the values of the params hash.

My approach would be to build a hash, and then run this:

``` ruby
query_params = # ... build hash
query_params.map { |(key, value| "#{CGI.escape(key)}=#{CGI.escape(value)}" }.join('&') 
```

lib/active_shipping/carriers/correios.rb
```diff
@@ -0,0 +1,271 @@
...
+
+      def create_url(package)
+        "#{URL}?#{query_string(params(package))}"
+      end
+
+      def service_type
+        @options[:services].nil? ? '41106,40010' : @options[:services].join(',')
```
wvanbergen: Maybe use a constant for this default value, so it's a bit more clear what it means?

lib/active_shipping/carriers/correios.rb
```diff
@@ -0,0 +1,271 @@
...
+          service_type: service_type,
+          origin_zip: origin.zip,
+          destination_zip: destination.zip,
+          mao_propria_extra: parse_boolean(options[:mao_propria_extra]),
+          declared_value_extra: parse_currency(options[:declared_value_extra]),
+          delivery_notice_extra: parse_boolean(options[:delivery_notice_extra]),
+          return_type: RETURN_TYPE,
+          return_information: RETURN_INFORMATION_TYPE[:prices]
+        }
+        @urls = packages.map { |package| create_url(package) }
```
wvanbergen: There's no way of doing a single request for the entire shipment?

lib/active_shipping/carriers/correios.rb
```diff
@@ -0,0 +1,271 @@
...
+
+      def initialize(request, raw_xmls)
+        @request = request
+        @raw_xmls = raw_xmls
+        @documents = raw_xmls.map { |xml| Nokogiri::XML(xml) }
+      end
+
+      def rate_response
+        @rates = rates
+        CorreiosRateResponse.new(true, 'success', params_options, response_options)
```
wvanbergen: Instead of `'success'`, you can just sue `nil` as the message parameter.

test/remote/correios_test.rb
```diff
@@ -0,0 +1,81 @@
...
+  def test_response_prices
+    response = @carrier.find_rates(@saopaulo, @riodejaneiro, [@book, @poster])
+    response_book= @carrier.find_rates(@saopaulo, @riodejaneiro, [@book])
+    response_poster = @carrier.find_rates(@saopaulo, @riodejaneiro, [@poster])
+
+    book_price = response_book.rates.sum(&:price)
+    poster_price = response_poster.rates.sum(&:price)
+    total_price = response.rates.sum(&:price)
+
+    assert total_price == (book_price + poster_price)
```
wvanbergen: `assert_equal total_price, book_price + poster_price`

test/remote/correios_test.rb
```diff
@@ -0,0 +1,81 @@
...
+  end
+
+  def test_invalid_zip
+    assert_raises(ActiveShipping::ResponseError) do
+      @carrier.find_rates(@saopaulo, @invalid_city, [@book])
+    end
+  rescue => error
+    assert_equal ActiveShipping::ResponseError, error.class
+    assert_equal RateResponse, error.response
+    assert_not error.message.empty?
```
wvanbergen: nitpick: `refute` instead of `assert_not`

wvanbergen:
Just some small comments, but this looks good!

albiere:
Thanks for pointing out some smells.
Well, the Correios webservice supports only one package per request. Basically, my implementation requests the webservice for each package passed in to `find_rates`. The final price is obtained by the sum of each package price. I don't know whether it's the best approach (perhaps calculating a new dimension based on all packages passed in). I read some other carrier implementations and they do almost the same thing.

Thanks!

lib/active_shipping/carriers/correios.rb
```diff
@@ -0,0 +1,276 @@
...
+        !text.empty? && text != "0"
+      end
+
+      def error(xml_item)
+        xml_item.css('Erro').text
+      end
+
+      def error_message(xml_item)
+        xml_item.css('MsgErro').text
+      end
```
wvanbergen: What is the difference between `error?`, `error`, and error_message`? Maybe the naming can be improved here.

wvanbergen:
Can you remove the `tags` file? I assume that one was committed by mistake.

test/unit/carriers/correios_test.rb
```diff
@@ -0,0 +1,160 @@
...
+    @carrier.stubs(:perform).returns([@response_book_invalid])
+    @carrier.find_rates(@saopaulo, @patosdeminas, [@book])
+
+    refute true, 'Must raise ActiveShipping::ResponseError'
+  rescue => error
+    assert_kind_of ActiveShipping::ResponseError, error
+    assert_equal "CEP de origem invalido", error.message
+    assert_equal error.response.raw_responses, [@response_book_invalid]
+    assert_equal Hash.new, error.response.params
+  end
```
wvanbergen:
```ruby 
error = assert_raises(ActiveShipping::ResponseError) do
  @carrier.stubs(:perform).returns([@response_book_invalid])
  @carrier.find_rates(@saopaulo, @patosdeminas, [@book])
end

assert_equal "CEP de origem invalido", error.message
assert_equal error.response.raw_responses, [@response_book_invalid]
assert_equal Hash.new, error.response.params
```

wvanbergen: Also, can you add a test where for a `find_rates` call with two packages, in which the first request succeeds and the second requests fails? I think the best way to handle that is to return an error response in that case.

test/remote/correios_test.rb
```diff
@@ -0,0 +1,75 @@
...
+    @carrier.find_rates(@saopaulo, @invalid_city, [@book])
+
+    refute true, 'Must raise ActiveShipping::ResponseError'
+  rescue => error
+    assert_kind_of ActiveShipping::ResponseError, error
+    assert_kind_of RateResponse, error.response
+    refute error.message.empty?
+    assert error.response.raw_responses.any?
+    assert_equal Hash.new, error.response.params
+  end
```
wvanbergen: Use `assert_raises` (see below)

test/remote/correios_test.rb
```diff
@@ -0,0 +1,75 @@
+require 'test_helper'
+
+class RemoteCorreiosTest < Minitest::Test
+  include ActiveShipping::Test::Fixtures
```
wvanbergen: I think you don't need this module anymore.

wvanbergen:
Some final nitpicks before I can merge. Thanks for being so responsive :)

maartenvg:
@celsodantas

lib/active_shipping/carriers/correios.rb
celsodantas: `error` is the error code. `error_message` is what is it.

From Correios docs:
screen shot 2015-02-24 at 4 41 06 pm

The XML contains both in different tags.

I agree that the naming might be improved. Maybe rename `error` to `error_id`.

albiere: Sure! I didn't like too much the name `error` either. I'll change it to `error_id` definitely! Thanks!

albiere:
Thank so much for helping me out! I knew there was a better way to use assert_raises.. hahaha
Let me know if you have any other doubt.

wvanbergen:
I'm happy now. @robfoster what do you think?

ghonorato:
+1

wvanbergen:
@RichardBlair @kmcphillips last look?

lib/active_shipping/carriers/correios.rb
```diff
@@ -0,0 +1,276 @@
...
+
+      def create_response(raw_xmls)
+        correios_response = CorreiosResponse.new(self, raw_xmls)
+        correios_response.rate_response
+      end
+
+      private
+
+      def parse_boolean(param)
+        param == true ? 'S' : 'N'
```
kmcphillips: The `== true` is unnecessary.

```ruby
param ? 'S' : 'N'
```

lib/active_shipping/carriers/correios.rb
```diff
@@ -0,0 +1,276 @@
...
+        }
+      end
+
+      def query_string(package)
+        params = build_params(package)
+        hash_query_string(params).map { |key, value| "#{CGI.escape(key)}=#{CGI.escape(value.to_s)}" }.join('&')
+      end
+
+      def create_url(package)
+        "#{URL}?#{query_string(package)}"
```
kmcphillips: Consider using something like `URI::HTTP.build`?

kmcphillips: Means you don't have to do the `join("&")` and `CGI.escape` stuff above too.

kmcphillips:
Looks good. A few comments. Nice use of small objects. Good test coverage. :+1:

wvanbergen:
@albiere if you have some time address the final comments, we should be able to merge this soon :+1:

lib/active_shipping/carriers/correios.rb
albiere: Cool! I'm working on it. it's a good point! thanks!

albiere:
@wvanbergen Just to let you know I'm working on it! Taking a bit too long because I'm refactoring the tests that make sure the URL is generated correctly. They are too brittle.

wvanbergen:
Cool, just make sure to leave me a comment here once you're done so I get a notification.

albiere:
@wvanbergen I made the changes. Have a look at the code and in case you find something wrong just let me know. (there's something wrong with the release, probably related to #249)

wvanbergen:
Thank you! The CI failures are unrelated, so I am merging.

PR #202, opened 2014-11-24
UPS needs shipment Description for cross-border

jamesroscoe:
UPS requires shipment description in order to deliver a product from Canada to the US (and probably any time customs clearance is required). Pass as {description: "description"} in the options hash to Ups#create_shipment.

PR #246, opened 2015-02-24
UPS. Better support for delivery confirmation options

eboswort:
I pulled this out of #202 to provide support for all of UPS's delivery confirmation options.

I translated it to nokogiri and added some logic to deal with shipment-level vs package-level delivery confirmation stuff. 
UPS specifies that for certain origin/destination pairs delivery_confirmation options must be specified on the shipment-level, for other origin/destination pairs that option must be specified on each package in the shipment.
As it's written a user can pass in delivery_confirmation as an option and it will be used shipment-level or applied to each package based on the origin/destination pair. If the user would like to specify different delivery_confirmation options for each package in the shipment she can do so by specifying a value in the package.options hash for each package.

It looks like I changed more than I actually did because `build_international_shipment_request_options` had to be split apart so that the ShipmentServiceOptions node could be used outside of that method.

### DIFF
--- a/lib/active_shipping/carriers/ups.rb
+++ b/lib/active_shipping/carriers/ups.rb
@@ -107,6 +107,18 @@ module ActiveShipping
     DEFAULT_SERVICE_NAME_TO_CODE['UPS 2nd Day Air'] = "02"
     DEFAULT_SERVICE_NAME_TO_CODE['UPS 3 Day Select'] = "12"
 
+    SHIPMENT_DELIVERY_CONFIRMATION_CODES = {
+      delivery_confirmation_signature_required: 1,
+      delivery_confirmation_adult_signature_required: 2
+    }
+
+    PACKAGE_DELIVERY_CONFIRMATION_CODES = {
+      delivery_confirmation: 1,
+      delivery_confirmation_signature_required: 2,
+      delivery_confirmation_adult_signature_required: 3,
+      usps_delivery_confirmation: 4
+    }
+
     def requirements
       [:key, :login, :password]
     end
@@ -361,6 +373,14 @@ module ActiveShipping
               end
             end
 
+            shipment_service_options = XmlNode.new('ShipmentServiceOptions') do |opts|
+              if delivery_confirmation = options[:delivery_confirmation]
+                opts << XmlNode.new('DeliveryConfirmation') do |dc|
+                  dc << XmlNode.new('DCISType', SHIPMENT_DELIVERY_CONFIRMATION_CODES[delivery_confirmation])
+                end
+              end
+            end
+
             if options[:international]
               build_international_shipment_request_options(xml, origin, destination, packages, options)
             end
@@ -579,14 +599,12 @@ module ActiveShipping
 
         unless options[:international]
           xml.PackageServiceOptions do
-            if package.options[:signature_required]
-              # Package level delivery confirmation is only available when shipping US -> US or PR -> PR
-              xml.DeliveryConfirmation do
-                xml.DCISType(2) # 2 = signature required.
+            if delivery_confirmation = options[:delivery_confirmation]
+              package_service_options << XmlNode.new('DeliveryConfirmation') do |dc|
+                dc << XmlNode.new('DCISType', PACKAGE_DELIVERY_CONFIRMATION_CODES[delivery_confirmation])
               end
-            elsif
-              xml.ShipperReleaseIndicator
             end
+            package_node << package_service_options if package_service_options.children.count > 0
           end
         end
 
--- a/lib/active_shipping/carriers/ups.rb
+++ b/lib/active_shipping/carriers/ups.rb
@@ -286,10 +286,12 @@ module ActiveShipping
     # * reference_numbers: Array of hashes with :value => a reference number value and optionally :code => reference number type
     # * prepay: if truthy the shipper will be bill immediatly. Otherwise the shipper is billed when the label is used.
     # * negotiated_rates: if truthy negotiated rates will be requested from ups. Only valid if shipper account has negotiated rates.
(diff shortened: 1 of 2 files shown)


eboswort:
Any comments on this one? I've got another PR I'd like to put up, but would like to get this merged first.

wvanbergen:
@Shopify/shipping

lib/active_shipping/carriers/ups.rb
```diff
@@ -835,5 +855,48 @@ def allow_package_level_reference_numbers(origin, destination)
...
+
+    # For certain origin/destination pairs, UPS allows each package in a shipment to have a specified delivery_confirmation option
+    # otherwise the delivery_confirmation option must be specified on the entire shipment.
+    # See Appendix P of UPS Shipping Package XML Developers Guide for the rules on which the logic below is based.
+    def package_level_delivery_confirmation?(origin, destination)
+      origin.country_code == destination.country_code ||
+      [['US','PR'], ['PR','US']].include?([origin,destination].map(&:country_code))
+    end
+
+    def assert_valid_package_level_delivery_confirmation
```
kmcphillips: This is empty and unused.

kmcphillips:
LGTM. Thanks for adding tests.

lib/active_shipping/carriers/ups.rb
eboswort: Good call. I'll remove it.

eboswort:
Just pushed a new version with the empty method that @kmcphillips found removed.

kmcphillips:
:+1:

lib/active_shipping/carriers/ups.rb
```diff
@@ -362,7 +378,32 @@ def build_shipment_request(origin, destination, packages, options={})
...
+            end
+
+            xml.ShipmentServiceOptions do
+              if delivery_confirmation = options[:delivery_confirmation]
+                xml.DeliveryConfirmation do
+                  xml.DCISType(SHIPMENT_DELIVERY_CONFIRMATION_CODES[delivery_confirmation])
+                end
+              end
+
+              if options[:international]
```
RichardBlair: I personally would prefer this node to be called `:customs`. It's a little more inline with that I've seen on other carriers.

lib/active_shipping/carriers/ups.rb
```diff
@@ -411,51 +452,35 @@ def build_delivery_dates_request(origin, destination, packages, pickup_date, opt
...
-            xml.FormType('01') # 01 is "Invoice"
-            xml.InvoiceDate(options[:invoice_date] || Date.today.strftime('%Y%m%d'))
-            xml.ReasonForExport(options[:reason_for_export] || 'SALE')
-            xml.CurrencyCode(options[:currency_code] || 'USD')
-
-            if options[:terms_of_shipment]
-              xml.TermsOfShipment(options[:terms_of_shipment])
-            end
+        xml.InternationalForms do
+          xml.FormType('01') # 01 is "Invoice"
```
RichardBlair: Should form type be something that is passed down through the options hash? Having this hard coded doesn't feel exactly right to me.

lib/active_shipping/carriers/ups.rb
```diff
@@ -411,51 +452,35 @@ def build_delivery_dates_request(origin, destination, packages, pickup_date, opt
...
-                  xml.Number((package.options[:item_count] || 1))
-                  xml.UnitOfMeasurement do |xml|
-                    # NMB = number. You can specify units in barrels, boxes, etc. Codes are in the api docs.
-                    xml.Code(package.options[:unit_of_item_count] || 'NMB')
-                  end
+          packages.each do |package|
+            xml.Product do |xml|
+              xml.Description(package.options[:description])
+              xml.CommodityCode(package.options[:commodity_code])
+              xml.OriginCountryCode(origin.country_code(:alpha2))
```
RichardBlair: It is my understanding that in customs forms the origin country is the country of origin for the item being shipped.

lib/active_shipping/carriers/ups.rb
```diff
@@ -411,51 +452,35 @@ def build_delivery_dates_request(origin, destination, packages, pickup_date, opt
...
-                  xml.Value(package.value / (package.options[:item_count] || 1))
-                  xml.Number((package.options[:item_count] || 1))
-                  xml.UnitOfMeasurement do |xml|
-                    # NMB = number. You can specify units in barrels, boxes, etc. Codes are in the api docs.
-                    xml.Code(package.options[:unit_of_item_count] || 'NMB')
-                  end
+          packages.each do |package|
+            xml.Product do |xml|
+              xml.Description(package.options[:description])
+              xml.CommodityCode(package.options[:commodity_code])
```
RichardBlair: Is `CommodityCode` the harmonized system code? If so, can we call it that in the options hash?

eboswort:
@RichardBlair 
About your last 3 comments, I feel like those are all valid points, but this PR is just moving the code that relates to customs. I needed to add something to the ShipmentServiceOptions node which was previously being created inside the `build_international_shipment_request_options ` method, so I pulled that out. I think any cleaning up of the support for customs should be done in a different PR.

#202 provides a potential structure for creating international forms which seems like an improvement on this, although it also has a bunch of other unrelated stuff in it and needs a rebase.

RichardBlair:
@eboswort if this is a refactor then disregard my comments. Sorry for the delay, I was away!

eboswort:
Shall we merge? @Shopify/shipping

eboswort:
@wvanbergen @kmcphillips @RichardBlair 
Sorry to be a pest, but seems like we're all set here right?

kmcphillips:
:+1:

wvanbergen:
:shipit: if remote tests are passing.

PR #249, opened 2015-04-01
Prepare for release of v1.0.0

wvanbergen:
@Shopify/shipping Let's get this out of the door so people stop complaining about the latest ActiveMerchant and ActiveShipping causing naming conflicts.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,22 @@
 # ActiveShipping CHANGELOG
 
+### v1.0.0
+
+- **BREAKING CHANGE:** Change namespace from `ActiveMerchant::Shipping` to `ActiveShipping`
+- Drop support for Ruby 1.8 and Ruby 1.9.
+- Drop support for ActiveSupport < 3.2, support up to ActiveSupport 4.2.
+- Updated Fedex to use latest API version for tracking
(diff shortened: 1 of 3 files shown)


wvanbergen:
Hmmm, looks like some Canada Post PWS API calls require a contract number all of the sudden (even when we're using the non-contract shipping API)

kmcphillips:
Would be worth putting a gem post install message for the breaking change.

In the `.gemfile`:

```ruby
gem.post_install_message "Warning breaking change etc. link to docs whatever"
```

jnormore:
> Hmmm, looks like some Canada Post PWS API calls require a contract number all of the sudden (even when we're using the non-contract shipping API)

https://www.canadapost.ca/cpo/mc/business/productsservices/developers/messagescodetables.jsf

It needs the ``platform_id`` to do ``CreateShipment`` requests on behalf of non contract customers.

wvanbergen:
@jnormore Hmm, they must have changes this since I worked on this 2 months ago, because those tests were working fine.

@kmcphillips I don't think the `post_install_message` is strictly necessary; backwards incompatible changes kind of comes with the territory of major releases, but we could include it.

kmcphillips:
That's fine, I don't feel strongly about it.

PR #250, opened 2015-04-09
Fix tests and add single service rates to UPS

hiemanshu:
Fixed some issues I found with how the delivery_dates tests were written. UPS would fail with an invalid pickup date exception. I also added support to get rates for only a single service from UPS.

### DIFF
--- a/lib/active_shipping/carriers/ups.rb
+++ b/lib/active_shipping/carriers/ups.rb
@@ -221,9 +221,7 @@ module ActiveShipping
         xml.RatingServiceSelectionRequest do
           xml.Request do
             xml.RequestAction('Rate')
-            xml.RequestOption('Shop')
-            # not implemented: 'Rate' RequestOption to specify a single service query
-            # xml.RequestOption((options[:service].nil? or options[:service] == :all) ? 'Shop' : 'Rate')
+            xml.RequestOption((options[:service].nil?) ? 'Shop' : 'Rate')
           end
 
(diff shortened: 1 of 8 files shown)


wvanbergen:
This LGTM: @kmcphillips @RichardBlair ?

wvanbergen:
Test failures are related to Canada Post :(

lib/active_shipping/carriers/ups.rb
```diff
@@ -253,6 +251,12 @@ def build_rate_request(origin, destination, packages, options = {})
             #                   * Shipment/AlternateDeliveryTime element
             #                   * Shipment/DocumentsOnly element
 
+            unless options[:service].nil?
```
kmcphillips: Ok with empty string here then?

kmcphillips:
Change looks good, though I am dubious that depending on `business_time` to have the exact same business days logic as the carrier may lead to some intermittent remote test failures.

eboswort:
I think business_time probably has a lot more holidays than UPS, since they only have a few:
http://www.ups.com/content/us/en/resources/ship/imp_exp/operation.html
But business_time is definitely an improvement on the current state of the ups tests as I wrote them (whoops!)

eboswort:
Can we merge so I can rebase #253 and get that merged :)

eboswort:
@wvanbergen Hey, can you pull the trigger on this one? ;)

kmcphillips:
Pulled

PR #256, opened 2015-05-06
Update README.md

abrahamatempireone:
Taken directly from http://shopify.github.io/active_shipping/


wvanbergen:
It's actually http://shopify.github.io/active_shipping/ that is outdated (it was generated based on an old version of this README). However, the new 1.0 version is only released as a prerelease.

wvanbergen:
I released 1.0 final, and updated http://shopify.github.io/active_shipping/

PR #259, opened 2015-05-11
Add repodb.yml

kmcphillips:
@Shopify/shipping


garethson:
:+1: Looks like build error is a timeout.

PR #260, opened 2015-05-13
New USPS tracking API

trishume:
This PR switches USPS package tracking to a new version of their API that offers additional functionality and a cleaner structure. See [here](https://www.usps.com/business/web-tools-apis/track-and-confirm.pdf) for documentation on the old and new APIs.

Switching to this API fixes 2 bugs:
- Tracking a number with only one event threw that event's message as an error even if nothing was wrong.
- Tracking a number that was a duplicate was impossible, the new API allows passing through disambiguation information.

It also adds support for 2 more pieces of information:
- The estimated time of delivery
- The USPS event codes. These are important because they are [documented](https://about.usps.com/publications/pub97/pub97_i.htm) which means they can be reliably used to determine if a package really has reached its destination and to reliably handle all the failure modes, even if the messages for them haven't yet been observed.

@Shopify/shipping


kmcphillips:
Can you please bump the gem version according to [semver](http://semver.org/). This is a functional change.

kmcphillips:
The `destination_zip` and `mailing_date` are optional, which is great. But will USPS still provide results if they are missing, assuming the tracking number is unique?

lib/active_shipping/carriers/usps.rb
```diff
@@ -232,26 +229,44 @@ def maximum_weight
...
+      EventDetails.new(description, time, zoneless_time, location, event_code)
     end
 
     protected
 
     def build_tracking_request(tracking_number, options = {})
       xml_builder = Nokogiri::XML::Builder.new do |xml|
-        xml.TrackRequest('USERID' => @options[:login]) do
-          xml.TrackID('ID' => tracking_number)
+        xml.TrackFieldRequest('USERID' => @options[:login]) do
```
kmcphillips: To be clear, `TrackFieldRequest` is what specifies the new API.

trishume:
Yes USPS will still provide results without them if it is unique. Luckily because otherwise it wouldn't be backwards compatible, I was super relieved when I saw 'optional'.

kmcphillips:
This is all great work. :+1:

lib/active_shipping/carriers/usps.rb
trishume: Yup, along with the revision tag to specify the newest new API. The current API that is being used is two versions old.

lib/active_shipping/carriers/usps.rb
```diff
@@ -232,26 +229,44 @@ def maximum_weight
...
+          xml.Revision { xml.text('1') }
+          xml.ClientIp { xml.text(@options[:client_ip] || '127.0.0.1') }
+          xml.SourceId { xml.text(@options[:source_id] || 'active_shipping') }
```
trishume: It is undocumented what this and the client IP are really supposed to mean. They seem to do nothing. Probably for internal analytics I guess. They're mandatory.

PR #261, opened 2015-05-14
Fedex Label Creation

eboswort:
This is a starting place for creating Fedex Labels. It provides a very small subset of the options available but it's working successfully to create labels in production for me.


lib/active_shipping/carriers/fedex.rb
```diff
@@ -153,8 +161,131 @@ def find_tracking_info(tracking_number, options = {})
...
+      begin
+        request = build_shipment_request(origin, destination, packages, options)
+        logger.debug(request) if logger
+
+        logger.debug(confirm_response) if logger
+
+        response = commit(save_request(request), (options[:test] || false))
+        parse_ship_response(response)
+      rescue RuntimeError => e
+        raise "Could not obtain shipping label. #{e.message}."
```
trishume: Under what circumstances would this happen in practice? If this should only happen in the case of a bug in `active_shipping` why bother wrapping the message? Why `RuntimeError`, was there a specific case you were thinking of with this?

With this you lose the class information of the actual error, and the resulting raised exception is not even an activeshipping error class.

lib/active_shipping/carriers/fedex.rb
```diff
@@ -153,8 +161,131 @@ def find_tracking_info(tracking_number, options = {})
...
+      end
+    end
+
     protected
 
+    def build_shipment_request(origin, destination, packages, options = {})
+      imperial = %w(US LR MM).include?(origin.country_code(:alpha2))
```
trishume: Maybe extract this logic into a helper instead of having it copy-pasted many times in the file.

lib/active_shipping/label.rb
```diff
@@ -0,0 +1,10 @@
+module ActiveShipping
+  class Label
+    attr_reader :tracking_number, :base64_img_data
+
+    def initialize(tracking_number, base64_img_data)
+      @tracking_number = tracking_number
+      @base64_img_data = base64_img_data
```
trishume: Storing arbitrary binary in Ruby strings is totally ok, in fact it is the preferred method of storage for random binary. You can use the standard library `require "base64"` module to decode the responses into raw format.

That way it will be easier to deal with in ways like directly writing it to a file and it doesn't unnecessarily expose the implementation details of the UPS and Fedex APIs.

trishume:
Awesome, thanks for contributing such a big and important feature! :+1: 

Just had a few minor comments, otherwise this is great stuff.

lib/active_shipping/carriers/fedex.rb
eboswort: This part was @jpsilvashy, but I think he pulled this structure from the UPS code which does the same thing.
I don't think it has a purpose. I could take it out. Do you think we should do the same for ups?

trishume: Yah we probably should remove it from both places. Although if we are lucky pinging @jmorton (guy who wrote it according to git blame) will lend some insight.

jmorton: Yikes! That was some pretty vague (and sloppy) exception handling on my part, sorry about that.  At a minimum, I should have used `ActiveShipping::Error`.

If I recall correctly, the combination of origin, destination, and shipping priority can be invalid.  In this case, the carrier API indicates and error.

trishume: Thanks @jmorton! This and the UPS code can probably be removed then. The base Response class will raise an ActiveShipping::ResponseError when the success variable is false. That plus the exception originating from the `create_shipment` function should be enough context.

eboswort:
@trishume 
I think I took care of the stuff you brought up. I took out the useless begin/rescue block. I changed it so that img data is base64 decoded before being stored. And I made a little helper method for that imperial check.

lib/active_shipping/carriers/fedex.rb
```diff
@@ -153,10 +161,129 @@ def find_tracking_info(tracking_number, options = {})
       parse_tracking_response(xml, options)
     end
 
+
+    # Get Shipping labels
+    # Caveats:
+    #  - Only supports singlular packages
```
kmcphillips: What is the behaviour when multiple packages are passed in then? Since the signature implies _not_ singular package support.

kmcphillips:
:+1:

lib/active_shipping/carriers/fedex.rb
trishume: Looking at the code it seems that some carriers use `packages` and some us `package` for this method, which isn't nice. The spec in `carriers.rb` does specify an array though. I was going to suggest just using `map` to handle multiple packages but then you don't get a single response like other methods.

Not sure if it is better to use a consistent ruby API with a comment caveat or use the inconsistent ruby API with no comment caveat.

trishume:
Great :+1:

I'm adding another new feature quite soon so when I'm done that and you're happy with this I can release both of them in a `1.2.0` release.

kmcphillips:
> release both of them in a 1.2.0 release.

+1

eboswort:
Awesome!

lib/active_shipping/carriers/fedex.rb
jmorton: I'm not intimately familiar with the UPS or FedEx APIs anymore, but I seem to recall that a single UPS shipment supports the notion of a single tracking number for a set of packages.  I'm pretty sure `map` would work in practice and is arguably more elegant, but it could leading to billing problems if discounts have been negotiated for multi-package shipments.

eboswort: Yes UPS definitely does support this. They have the concept of a Shipment that contains multiple Packages. I think that Fedex supports it as well, but this is one of many features I didn't have time to implement.
We could replace the comment with an error that says the functionality is not implemented.

trishume:
I released this in `v1.2.0` now on Rubygems. Thanks for the great work!

PR #262, opened 2015-05-15
Fix USPS tracking bug

trishume:
In #260 I assumed some fields would always be present that are in fact optional. This caused errors when for example, tracking a number with no predicted delivery date yet.

This PR passes through nil values for those fields instead of throwing an error. Only the predicted delivery dates have been observed to be missing but this PR also speculatively allows the location to be optional since some more obscure events like "seized by law enforcement" may not have a location. 

@Shopify/shipping

### DIFF
--- a/lib/active_shipping/carriers/usps.rb
+++ b/lib/active_shipping/carriers/usps.rb
@@ -238,9 +238,9 @@ module ActiveShipping
       
       timestamp = "#{node.at('EventDate').text}, #{node.at('EventTime').text}"
       event_code = node.at('EventCode').text
-      city = node.at('EventCity').text
-      state = node.at('EventState').text
-      zip_code = node.at('EventZIPCode').text
+      city = try_text(node.at('EventCity'))
+      state = try_text(node.at('EventState'))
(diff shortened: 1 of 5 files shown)


lib/active_shipping/carriers/usps.rb
```diff
@@ -238,9 +238,9 @@ def extract_event_details(node)
       
       timestamp = "#{node.at('EventDate').text}, #{node.at('EventTime').text}"
       event_code = node.at('EventCode').text
-      city = node.at('EventCity').text
-      state = node.at('EventState').text
-      zip_code = node.at('EventZIPCode').text
+      city = try_text(node.at('EventCity'))
```
eboswort: Can't you just do:
`node.at('EventCity').try(:text)`

I don't think you need to define the `try_text` method

trishume: I wasn't sure that method was defined since this isn't a rails app. TIL active_shipping depends on activesupport. Thanks!

kmcphillips:
Change all looks good.

eboswort:
LGTM as well!

PR #265, opened 2015-05-22
add reference for estimates

RichardBlair:
This adds a field for rate estimate references for services that guarantee quotes.

Review @kmcphillips @j-mutter


kmcphillips:
Can you include a version bump?

kmcphillips:
Version bump is on master already.

PR #266, opened 2015-05-25
Fix finding of error description in USPS tracking

trishume:
The existing code for error message extraction wasn't tested, and was totally wrong.

#### Code Issues Fixed in this PR

- `document.elements[...]` doesn't search the document, it indexes an array.
- failed searches return an empty array-ish-thing not nil so `first_thing || second_thing` doesn't work
- Same thing made the checking multiple xpaths pattern not work, replaced it with a single general xpath.

I'm ashamed that I didn't have to Google the xpath syntax. I try to repress the memories of my days working with XSLT.

@Shopify/shipping

### DIFF
--- a/lib/active_shipping/carriers/usps.rb
+++ b/lib/active_shipping/carriers/usps.rb
@@ -154,11 +154,6 @@ module ActiveShipping
       "WS" => "Western Samoa"
     }
 
-    STATUS_NODE_PATTERNS = %w(
-      Error/Description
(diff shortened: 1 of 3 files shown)


trishume:
Failing tests are just `RemoteCorreiosTest` as usual.

kmcphillips:
Well I do have to google xpath. :+1:

RichardBlair:
:+1:

PR #267, opened 2015-05-27
UPS - add tracking_option on tracking info request

alvingarcia:
In order to track a UPS Mail Innovations tracking number in the API, you must also include the TrackingOption which is 03.

Here is an example of what the request would look like:

```xml
<?xml version="1.0" encoding="UTF-8"?>
<AccessRequest xml:lang="en-US">
<AccessLicenseNumber>Your Access Key</AccessLicenseNumber>
<UserId>Your User ID</UserId>
<Password>Your Password</Password>
</AccessRequest>
<?xml version="1.0" encoding="UTF-8"?>
<TrackRequest>
<TrackingOption>03</TrackingOption>
<Request>
<TransactionReference/>
</Request>
<RequestAction>Track</RequestAction>
<RequestOption>1</RequestOption>
<TrackingNumber>1234567890</TrackingNumber>
</TrackRequest>
```

### DIFF
--- a/lib/active_shipping/carriers/ups.rb
+++ b/lib/active_shipping/carriers/ups.rb
@@ -514,6 +514,7 @@ module ActiveShipping
     def build_tracking_request(tracking_number, options = {})
       xml_builder = Nokogiri::XML::Builder.new do |xml|
         xml.TrackRequest do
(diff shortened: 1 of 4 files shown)


kmcphillips:
Thanks

PR #268, opened 2015-05-27
Handle ZIP+4 by trimming.

trishume:
Currently tracking API calls fail in an invalid-schema kind of way if the zip code format is invalid. [ZIP+4](https://en.wikipedia.org/wiki/ZIP_code#ZIP.2B4) is considered an invalid format. This PR trims the first 5 digits to turn a ZIP+4 into a valid normal ZIP.

@Shopify/shipping


trishume:
Just figured out there's already a method for stripping zip codes and that I hadn't written any test. Did that now.

kmcphillips:
Good find. :+1:

PR #269, opened 2015-05-27
Skip remote Correios tests, they are unreliable.

trishume:
Because travis tests 12 different versions and one of these tests fails I'd estimate around 10% of the time basic statistics tells us that these tests will break CI 70% of the time.

So we kill them :fire: 

@kmcphillips


kmcphillips:
Great solution to working around an unreliable API.

PR #270, opened 2015-05-28
Add support for batch tracking info fetching on USPS

trishume:
Adds a method `batch_find_tracking_info` that allows fetching the tracking info for multiple numbers at a time. It accepts an array of hashes with all the info necessary to uniquely specify a shipment and returns an array of results in the same order.

USPS has two types of errors, schema errors that wreck the whole request, which are thrown as exceptions, and individual errors that only affect a certain tracker, which are just returned as unsuccessful responses.

Both the single and batched methods use the same parsing and building methods behind the scenes.

@Shopify/shipping


kmcphillips:
> Both the single and batched methods use the same parsing and building methods behind the scenes.

kmcphillips:
Looks great!

PR #271, opened 2015-05-29
Error handling for FedEx unsupported multiple packages.

trishume:
As discussed in #261 the `create_shipment` interface accepts an array but the FedEx code doesn't support multiple packages yet. This just turns a comment caveat into an exception and then tests the interface.

@kmcphillips


kmcphillips:
Big improvement. Thanks @trishume.

cc @eboswort as this will be in the next release.

PR #272, opened 2015-05-31
Add OriginZip to USPS

kmcphillips:
@Shopify/shipping 

Pulls https://github.com/RichardBlair/active_shipping/pull/1 and https://github.com/RichardBlair/active_shipping/pull/2 into `master`.

Bumps version to `1.2.1`.

The change is described here in the USPS Web Tools API May release (warning: .RTF for some reason):
https://www.usps.com/business/web-tools-apis/2015-may-webtools-release-notes.rtf

### DIFF
--- a/lib/active_shipping/carriers/usps.rb
+++ b/lib/active_shipping/carriers/usps.rb
@@ -291,7 +291,7 @@ module ActiveShipping
     end
 
     def world_rates(origin, destination, packages, options = {})
-      request = build_world_rate_request(packages, destination, options)
+      request = build_world_rate_request(origin, packages, destination, options)
       # never use test mode; rate requests just won't work on test servers
       parse_rate_response(origin, destination, packages, commit(:world_rates, request, false), options)
     end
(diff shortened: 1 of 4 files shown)


lib/active_shipping/carriers/usps.rb
```diff
@@ -371,12 +371,15 @@ def build_us_rate_request(packages, origin_zip, destination_zip, options = {})
...
     # package.options[:mail_type] -- one of [:package, :postcard, :matter_for_the_blind, :envelope].
     #                                 Defaults to :package.
-    def build_world_rate_request(packages, destination, options)
+    def build_world_rate_request(origin, packages, destination, options)
       country = COUNTRY_NAME_CONVERSIONS[destination.country.code(:alpha2).value] || destination.country.name
       xml_builder = Nokogiri::XML::Builder.new do |xml|
         xml.IntlRateV2Request('USERID' => @options[:login]) do
           Array(packages).each_with_index do |package, id|
             xml.Package('ID' => id) do
+              if country == "Canada"
```
trishume: I bet everything works fine if you don't use an if statement, and that also makes it more robust to USPS adding additional countries to this requirement.

kmcphillips: I can get behind that.

lib/active_shipping/carriers/usps.rb
```diff
@@ -371,12 +371,13 @@ def build_us_rate_request(packages, origin_zip, destination_zip, options = {})
...
     # package.options[:mail_type] -- one of [:package, :postcard, :matter_for_the_blind, :envelope].
     #                                 Defaults to :package.
-    def build_world_rate_request(packages, destination, options)
+    def build_world_rate_request(origin, packages, destination, options)
       country = COUNTRY_NAME_CONVERSIONS[destination.country.code(:alpha2).value] || destination.country.name
       xml_builder = Nokogiri::XML::Builder.new do |xml|
         xml.IntlRateV2Request('USERID' => @options[:login]) do
           Array(packages).each_with_index do |package, id|
             xml.Package('ID' => id) do
+              xml.OriginZip(origin.zip)
```
RichardBlair: OriginZip is only required when the destination is canada atm.

trishume: I know it is but if it doesn't break things if we include it everywhere, why not?

My experience with USPS APIs suggest that it should be okay to include it everywhere and it will just ignore it when it isn't needed. I'm not sure though, it would need to be tested.

RichardBlair: Yes, :tophat: is required.

mleglise:
+1 for this patch.
Without it, Spree sites are returning $0 Priority Mail International shipping from the US to Canada.
How soon can we see this merged?

trishume:
This patch doesn't even fix it. I'll create a new PR with the real patch that we developed yesterday.

PR #273, opened 2015-06-01
Fix start_with method call

JaredMiller:
start_with rather than starts_with

### DIFF
--- a/lib/active_shipping/carriers/usps.rb
+++ b/lib/active_shipping/carriers/usps.rb
@@ -234,7 +234,7 @@ module ActiveShipping
     def extract_event_details(node)
       description = node.at('Event').text.upcase
 
-      if prefix = ONLY_PREFIX_EVENTS.find { |p| description.starts_with?(p) }
+      if prefix = ONLY_PREFIX_EVENTS.find { |p| description.start_with?(p) }
         description = prefix
       end
 


kmcphillips:
Thanks. I'm going to release `1.2.1` today so it will be included in that.

kmcphillips:
Introduced in 53c0e4d9ed2abc6fffc5bbe6d13189a1b928dc62 cc @trishume

kmcphillips:
So this is an `ActiveSupport` alias:
https://github.com/rails/rails/blob/08754f12e65a9ec79633a605e986d0f1ffa4b251/activesupport/lib/active_support/core_ext/string/starts_ends_with.rb#L2

Was this a bug for you? We have `ActiveSupport` as an explicit dependency.

JaredMiller:
@kmcphillips Yes, I was getting no method errors last night after upgrading from a very old version of active_shipping

It was failing on a very simple test file

    require 'json'
    require 'active_shipping'
    require 'yaml'
    include ActiveShipping
    
    loginConfig = YAML::load(File.open('<config file>'))
    usps_api = USPS.new(:login => loginConfig['usps']['login_name'])
    response = usps_api.find_tracking_info('9574211132094339557534')

kmcphillips:
Oh this is very interesting.

So we depend on active support but we don't actually require it anywhere. Since you're working outside of a `Gemfile` in the above example it's not being loaded automatically.

We'll change it so that we explicitly `require "active_support"`. @trishume has offered to PR that.

trishume:
https://github.com/Shopify/active_shipping/blob/master/lib/active_shipping.rb#L24 :neutral_face:

Now I'm even more confused. I have no idea why you were getting those errors.

trishume:
Ok at least I can reproduce it locally. Still no idea why this is happening though.

wvanbergen:
`require 'active_support'` doesn't actually load much. If you want to load all of AS, do `require 'active_support/all'`

wvanbergen:
If you include ActiveShipping in a Rails app, you won't likely run into this because Rails will load the necessary part of ActiveSupport.

trishume:
Thanks @wvanbergen that explains a lot. Now I know why it didn't also crash from the calls to `try`. I implemented this fix in #274.

PR #274, opened 2015-06-01
Require all of active_support.

trishume:
Fixes the root cause of the error encountered in #273.

@kmcphillips


kmcphillips:
Yup. This explains it. Thanks. :+1:

PR #276, opened 2015-06-02
Update USPS rates API - Fix errors

trishume:
This fixes the issues introduced by the USPS WEBTOOLS upgrade on May 31. This patch has been tested to work in retrieving international rates from the US to Canada.

Unlike the earlier attempt in #272 this patch passes format validation by including all the required fields and *having them in the correct order*. Also note that the `AcceptanceDateTime` field does not accept any time with a positive timezone, this is worked around by always converting to UTC.

@Shopify/shipping 
cc @mleglise because he noted that Spree was affected
cc @eboswort because Casper is on the list of Spree users and he might have encountered this

### DIFF
--- a/lib/active_shipping/carriers/usps.rb
+++ b/lib/active_shipping/carriers/usps.rb
@@ -404,7 +404,7 @@ module ActiveShipping
                 xml.public_send(COMMERCIAL_FLAG_NAME.fetch(commercial_type), 'Y')
               end
               xml.OriginZip(origin.zip)
-              xml.AcceptanceDateTime(DateTime.now.iso8601)
+              xml.AcceptanceDateTime(Time.now.utc.iso8601)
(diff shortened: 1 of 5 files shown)


kmcphillips:
Looks good. 

Can you bump the version too?

trishume:
Done

mleglise:
Thank you @trishume 
Is there any change of getting this backported on top of v0.12.5? There are plenty of production apps out there that need this fix, but cannot yet upgrade to the latest active_shipping.

trishume:
@mleglise done! Thanks for the backport!

For other people's reference we coordinated via email and @mleglise backported the fix in this PR on top of active_shipping v0.12.5

It is now in the `v0.12.6-hotfix` branch of this repo, in a git tag, and deployed to Rubygems:
https://rubygems.org/gems/active_shipping/versions/0.12.6

PR #277, opened 2015-06-03
Handle service specific USPS rate errors more gracefully

trishume:
When USPS changed their API the error being bubbled up was very unhelpful and not properly an active_shipping error. The reason for this is that the code for handling errors in rate response was missing a matcher for a specific type of error that was specific to a service.

@RichardBlair @kmcphillips

### DIFF
--- a/lib/active_shipping/carriers/usps.rb
+++ b/lib/active_shipping/carriers/usps.rb
@@ -416,7 +416,7 @@ module ActiveShipping
 
       xml = Nokogiri.XML(response)
 
-      if error = xml.at('/Error')
+      if error = xml.at_xpath('/Error | //ServiceErrors/ServiceError')
         success = false
         message = error.at('Description').text
       else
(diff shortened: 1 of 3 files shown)


lib/active_shipping/carriers/usps.rb
```diff
@@ -416,7 +416,7 @@ def parse_rate_response(origin, destination, packages, response, options = {})
 
       xml = Nokogiri.XML(response)
 
-      if error = xml.at('/Error')
+      if error = xml.at_xpath('/Error | //ServiceErrors/ServiceError')
```
kmcphillips: :+1:

kmcphillips:
This is super great. Thank you.

lib/active_shipping/carriers/usps.rb
RichardBlair: #xpath

Issue #278, opened 2015-06-04
FedEx Ground Home Delivery's rate estimate delivery dates are nil because they need to be parsed manually


ajsharma:
I think this similar to https://github.com/Shopify/active_shipping/issues/61. The XML for a `GROUND_HOME_DELIVERY` uses text for its transit time rather than a date.

Here's the relevant XML blob

```xml
<RateReplyDetails>
  <ServiceType>GROUND_HOME_DELIVERY</ServiceType>
  <PackagingType>YOUR_PACKAGING</PackagingType>
  <DeliveryStation>QNYA </DeliveryStation>
  <CommitDetails>
    <ServiceType>GROUND_HOME_DELIVERY</ServiceType>
    <TransitTime>THREE_DAYS</TransitTime>
    <BrokerToDestinationDays>0</BrokerToDestinationDays>
  </CommitDetails>
  <DestinationAirportId>EWR</DestinationAirportId>
  <IneligibleForMoneyBackGuarantee>false</IneligibleForMoneyBackGuarantee>
  <OriginServiceArea>A2</OriginServiceArea>
  <DestinationServiceArea>A1</DestinationServiceArea>
  <TransitTime>THREE_DAYS</TransitTime>
  <SignatureOption>SERVICE_DEFAULT</SignatureOption>
  <ActualRateType>PAYOR_ACCOUNT_PACKAGE</ActualRateType>
  <RatedShipmentDetails>
    <ShipmentRateDetail>
      <RateType>PAYOR_ACCOUNT_PACKAGE</RateType>
      <RateZone>6</RateZone>
      <RatedWeightMethod>ACTUAL</RatedWeightMethod>
      <MinimumChargeType>CUSTOMER</MinimumChargeType>
      <DimDivisor>0</DimDivisor>
      <FuelSurchargePercent>4.0</FuelSurchargePercent>
      <TotalBillingWeight>
        <Units>LB</Units>
        <Value>1.0</Value>
      </TotalBillingWeight>
      <TotalBaseCharge>
        <Currency>USD</Currency>
        <Amount>7.94</Amount>
      </TotalBaseCharge>
      <TotalFreightDiscounts>
        <Currency>USD</Currency>
        <Amount>1.33</Amount>
      </TotalFreightDiscounts>
      <TotalNetFreight>
        <Currency>USD</Currency>
        <Amount>6.61</Amount>
      </TotalNetFreight>
      <TotalSurcharges>
        <Currency>USD</Currency>
[...]

PR #279, opened 2015-06-04
Parse delivery date for FedEx Ground Home Delivery rate estimates

ajsharma:
Fixes https://github.com/Shopify/active_shipping/issues/278


kmcphillips:
Could we get some test coverage for this?

ajsharma:
@kmcphillips :+1: working on it now :smile:

ajsharma:
I've added a test for the behavior, hope it's enough :worried:

trishume:
:+1:

Not sure if that XML blob should be included in its entirety though. It's enormous. Might be useful reference but I think some redaction might be in order.

kmcphillips:
:+1: but also agree with possibly trimming the unused part of the XML, if possible.

ajsharma:
:+1: will do :)

ajsharma:
I removed all the shipping options besides the relevant ground shipping one. Is there anything else I should clean up?

trishume:
Looks good, I'll merge this somewhat soonish when I do the next release. Thanks for the contribution!

PR #248, opened 2015-03-17
USPS US to US rate requests must specify a container type

robfoster:
### Problem
When no container type is specified we set the container type to `<Container/>`. On large packages (packages with at least one dimension greater than 1 foot) the rate response comes back but is incorrect.

### Solution
Choose a container type when non is specified. 

### Notes
* A container type was being specified for international rate requests

@Shopify/shipping @wvanbergen

PR #280, opened 2015-06-05
Don't include container on small packages

### LINKED ISSUE #275: USPS - RECTANGULAR is an invalid container type for a REGULAR package and ALL service.
I'm getting the following error when trying to find_rates for USPS, any ideas?

     ActiveShipping::ResponseError:
       RECTANGULAR is an invalid container type for a REGULAR package and ALL service.

origin.to_s
=> "Lansing, MI, 48910 United States"
destination.to_s
=> "Tampa, FL, 33609 United States"
packages.to_s
=> "[#<ActiveShipping::Package:0x00000103f8ec98 @options={:units=>:imperial}, @dimensions=[#<Quantified::Length: 10 inches>, #<Quantified::Length: 10 inches>, #<Quantified::Length: 10 inches>], @weight_unit_system=:imperial, @dimensions_unit_system=:imperial, @weight=#<Quantified::Mass: 96.0 ounces>, @value=nil, @currency=nil, @cylinder=false, @gift=false, @oversized=false, @unpackaged=false, @inches=[10, 10, 10]>]"

trishume:
This PR fixes #275. According to the USPS documentation the container tag should not be included when the package isn't LARGE and it in fact gives an error. https://github.com/Shopify/active_shipping/pull/248 mistakenly added the container to every us to us rate request, causing them to fail.

This PR fixes a couple remote tests that failed because of this bug but nobody noticed because we don't have credentials in CI. 

I used credentials locally and ran the USPS remote tests, most are fixed, one still fails because of a different issue relating to packages with only a country (a rare case), I will fix that in a separate upcoming PR. I'll bump and release a new version after that PR.

@RichardBlair @kmcphillips cc @jnormore

### DIFF
--- a/lib/active_shipping/carriers/usps.rb
+++ b/lib/active_shipping/carriers/usps.rb
@@ -343,8 +343,11 @@ module ActiveShipping
               xml.ZipDestination(strip_zip(destination_zip))
               xml.Pounds(0)
               xml.Ounces("%0.1f" % [package.ounces, 1].max)
-              xml.Container(CONTAINERS[package.options[:container]] || (package.cylinder? ? 'NONRECTANGULAR' : 'RECTANGULAR'))
-              xml.Size(USPS.size_code_for(package))
+              size_code = USPS.size_code_for(package)
+              container = CONTAINERS[package.options[:container]]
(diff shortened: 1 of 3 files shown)


RichardBlair:
:shipit:

kmcphillips:
:+1:

PR #281, opened 2015-06-05
Fix USPS requests with only a country

trishume:
A failing remote test tipped me off to this. If the destination postal code was nil because only a country was provided then the API would return an error because AcceptanceDateTime is only allowed when DestinationPostalCode is present.

This PR fixes the remote test and also creates a local test for the fix.

After this and #280 are merged the USPS remote tests will all be passing and I will release a new bugfix release.

@RichardBlair @kmcphillips cc @jnormore

### DIFF
--- a/lib/active_shipping/carriers/usps.rb
+++ b/lib/active_shipping/carriers/usps.rb
@@ -404,8 +404,10 @@ module ActiveShipping
                 xml.public_send(COMMERCIAL_FLAG_NAME.fetch(commercial_type), 'Y')
               end
               xml.OriginZip(origin.zip)
-              xml.AcceptanceDateTime((options[:acceptance_time] || Time.now.utc).iso8601)
-              xml.DestinationPostalCode(destination.zip)
+              if destination.zip.present?
+                xml.AcceptanceDateTime((options[:acceptance_time] || Time.now.utc).iso8601)
+                xml.DestinationPostalCode(destination.zip)
+              end
             end
           end
         end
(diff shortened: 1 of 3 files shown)


lib/active_shipping/carriers/usps.rb
```diff
@@ -404,8 +404,11 @@ def build_world_rate_request(origin, packages, destination, options)
                 xml.public_send(COMMERCIAL_FLAG_NAME.fetch(commercial_type), 'Y')
               end
               xml.OriginZip(origin.zip)
-              xml.AcceptanceDateTime((options[:acceptance_time] || Time.now.utc).iso8601)
-              xml.DestinationPostalCode(destination.zip)
+              # AcceptanceDateTime without DestinationPostalCode isn't allowed
```
RichardBlair: Remove comment. The code explains this just fine.

RichardBlair:
other than the one comment, :shipit:

lib/active_shipping/carriers/usps.rb
```diff
@@ -404,8 +404,10 @@ def build_world_rate_request(origin, packages, destination, options)
                 xml.public_send(COMMERCIAL_FLAG_NAME.fetch(commercial_type), 'Y')
               end
               xml.OriginZip(origin.zip)
-              xml.AcceptanceDateTime((options[:acceptance_time] || Time.now.utc).iso8601)
-              xml.DestinationPostalCode(destination.zip)
+              if destination.zip
```
kmcphillips: Presence? Worried about empty string here?

kmcphillips:
:+1:

PR #282, opened 2015-06-05
Handle upper case country names in tracking

trishume:
This PR adds a remote test for USPS tracking using a tracking number of my own which I am okay with releasing. It also fixes a bug I found in the process of actually running this remote test which is that USPS returns upper case country names, which `ActiveUtils::Country` can't find. Thus this PR adds code for finding country names in a case insensitive way.

@kmcphillips @RichardBlair

### DIFF
--- a/lib/active_shipping/carriers/usps.rb
+++ b/lib/active_shipping/carriers/usps.rb
@@ -246,7 +246,9 @@ module ActiveShipping
 
       country_node = node.at('EventCountry')
       country = country_node ? country_node.text : ''
-      country = 'USA' if country.empty?
+      country = 'UNITED STATES' if country.empty?
+      # USPS returns upcased country names which ActiveUtils doesn't recognize without translation
+      country = find_country_code_case_insensitive(country)
 
       time = Time.parse(timestamp)
       zoneless_time = Time.utc(time.year, time.month, time.mday, time.hour, time.min, time.sec)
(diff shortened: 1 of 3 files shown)


kmcphillips:
How long to tracking numbers stay available online? It may disappear eventually. Which I guess is ok, I'm sure it's not short term.

trishume:
I dunno how long they last. I hope forever, they certainly have enough numbers to keep them around forever. It can always be replaced with a new one if it breaks.

kmcphillips:
I can agree with that line of thinking.

lib/active_shipping/carriers/usps.rb
```diff
@@ -642,6 +644,13 @@ def response_message(document)
       response_status_node(document).text
     end
 
+    def find_country_code_case_insensitive(name)
+      upcase_name = name.upcase
+      country = ActiveUtils::Country::COUNTRIES.detect { |c| c[:name].upcase == upcase_name }
+      raise ActiveShipping::Error, "No uppercase country found for #{name}" unless country
```
kmcphillips: I think the `uppercase` here is unnecessary.

kmcphillips:
:+1:

lib/active_shipping/carriers/usps.rb
trishume: I can remove that I guess. I wanted to distinguish it from the ActiveUtils error a bit but it is already an ActiveShipping error with a backtrace to this method.

PR #283, opened 2015-06-05
UPS. Support for voiding shipments.

eboswort:
This adds the ability to void UPS shipments, which is the same as canceling them. It's a good thing to do if you're not going to use a label. 

I'm wondering how the community feels about the way I set up the return values for `void_shipment`. The way I've got it, you get true if the shipment is voided and otherwise it raises an error with a message about what went wrong.

I guess I could return a response object but it seemed to me like a void shipment request is going to work 99% of the time and when it fails it's probably pretty unexpected and could merit an error rather than just a response with success? == false. 
But I'm happy to make a barebones response object if that's what y'all prefer.

Oh also, side note about the switch from `update` -> `merge`. In my fork of this repo we've switched all the `update`'s to `merge`'s. I don't think calling `update` makes sense because it makes changes to the carrier's `@options` instance variable. I would think if someone passes an options hash to a method then his or her intention would be to override those options just for the method call that he or she is performing right then, not for future calls of other methods on that carrier. So... I guess I'll put up a pr to switch them all?

### DIFF
--- a/lib/active_shipping/carriers/ups.rb
+++ b/lib/active_shipping/carriers/ups.rb
@@ -828,7 +828,7 @@ module ActiveShipping
       if success
         true
       else
-        raise "Void shipment failed with message: #{message}"
+        raise ResponseError.new("Void shipment failed with message: #{message}")
       end
     end
 
--- a/lib/active_shipping/carriers/ups.rb
+++ b/lib/active_shipping/carriers/ups.rb
@@ -16,7 +16,8 @@ module ActiveShipping
       :track => 'ups.app/xml/Track',
       :ship_confirm => 'ups.app/xml/ShipConfirm',
       :ship_accept => 'ups.app/xml/ShipAccept',
-      :delivery_dates =>  'ups.app/xml/TimeInTransit'
+      :delivery_dates =>  'ups.app/xml/TimeInTransit',
+      :void =>  'ups.app/xml/Void'
     }
 
     PICKUP_CODES = HashWithIndifferentAccess.new(
@@ -144,7 +145,7 @@ module ActiveShipping
     # @return [ActiveShipping::TrackingResponse] The response from the carrier. This
     #   response should a list of shipment tracking events if successful.
     def find_tracking_info(tracking_number, options = {})
-      options = @options.update(options)
+      options = @options.merge(options)
       access_request = build_access_request
       tracking_request = build_tracking_request(tracking_number, options)
       response = commit(:track, save_request(access_request + tracking_request), options[:test])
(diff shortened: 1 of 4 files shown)


lib/active_shipping/carriers/ups.rb
```diff
@@ -194,6 +195,14 @@ def get_delivery_date_estimates(origin, destination, packages, pickup_date=Date.
       parse_delivery_dates_response(origin, destination, packages, response, options)
     end
 
+    def void_shipment(tracking, options={})
+      options = @options.merge(options)
+      access_request = build_access_request
+      void_request = build_void_request(tracking)
+      response = commit(:void, save_request(access_request + void_request), (options[:test] || false))
```
kmcphillips: `(options[:test] || false)` can be replaced with `!!options[:test]`

kmcphillips: I also find the overloaded use of the `+` operator to be confusing. Is it concatenating two XML documents?

lib/active_shipping/carriers/ups.rb
```diff
@@ -800,6 +821,17 @@ def parse_delivery_dates_response(origin, destination, packages, response, optio
...
     end
 
+    def parse_void_response(response, options={})
+      xml = build_document(response, 'VoidShipmentResponse')
+      success = response_success?(xml)
+      message = response_message(xml)
+      if success
+        true
+      else
+        raise "Void shipment failed with message: #{message}"
```
kmcphillips: Either return a falsy response object, or raise a typed exception. You'll want to be able to catch it and perform an action. Raising a standard error makes it difficult to cleanly recover from.

eboswort:
@kmcphillips I took your advice on using a more specific error type. I saw there was already an ActiveShipping::ResponseError which seemed appropriate.

For your other two comments I'm just following the syntax used in every other call to `commit` in the ups.rb class. And the `(options[:test] || false)` thing is done in the fedex carrier as well.

I think you make valid points but it seems maybe someone could do a separate PR for coding style updates that does a mass replacement of `options[:test] || false` and concatenates the xml in a way that's a bit more clear. Because if I'm gonna change those things in this method then it would probably make sense to change them everywhere, which would make this a bit of a messy PR.

trishume:
At least one local test would be nice, we still don't run remote tests on CI. We're possibly going to get an internal Shopify CI with credentials soon but it's still nice for people and robots without credentials.

Other than that looks good! :+1: 

Agreed that the style can be a separate PR.

eboswort:
I added a simple local test. It would be kind of hard to break as far as tests go but I guess it's something.

trishume:
Great. Thanks again for being such a great contributor!

I'll merge this and ship it in the next release.

kmcphillips:
:+1: thanks

trishume:
Had to merge manually because of conflicts, they were trivial, just two methods added in the same place.

PR #286, opened 2015-06-08
Fix/skip remaining remote test failures

trishume:
This fixes a Stamps failure relating to their images now having a comment at the top. 

It also skips some Canada Post PWS tests that are failing but neither @wvanbergen or I can figure out how to fix them. They changed their API at some point but the docs haven't been updated to show the correct way, there is only a reference to the error [here](https://www.canadapost.ca/cpo/mc/business/productsservices/developers/messagescodetables.jsf). According to Willem it might have something to do with different account types. We agreed it is better to skip them for now than leave them failing as a disincentive to actually running the remote tests.

With this all the remote tests now either succeed or are skipped, given credentials.

@wvanbergen @kmcphillips 

screen shot 2015-06-08 at 4 29 54 pm

### DIFF
--- a/test/remote/stamps_test.rb
+++ b/test/remote/stamps_test.rb
@@ -143,7 +143,7 @@ class RemoteStampsTest < Minitest::Test
 
     assert_nil response.label_url
 
-    assert_equal "\r\nN\r\n", response.image[0..4]
+    assert_equal ";\r\n; ", response.image[0..4]
   end
 
   def test_international_shipment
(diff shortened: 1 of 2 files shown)


test/remote/stamps_test.rb
```diff
@@ -143,7 +143,7 @@ def test_shipment
 
     assert_nil response.label_url
 
-    assert_equal "\r\nN\r\n", response.image[0..4]
+    assert_equal ";\r\n; ", response.image[0..4]
```
kmcphillips: Intuitive!

kmcphillips:
:+1:

wvanbergen:
:shipit:

PR #288, opened 2015-06-12
UPS. BillThirdParty

nrevko:
##### PURPOSE:
This code adds ability to bill third party for UPS carrier when creating shipment.  
The user has to provide the following information:
- billing_account - UPS account number of the third party shipper
- billing_zip - postal code for the billing_account's pickup address 
- billing_country - country code for the billing_account's pickup address

##### NOTE: 
For a remote test for third party billing to run a UPS account for the third party shipper needs to be set up and added in the following environment variables: ACTIVESHIPPING_UPS_THIRD_PARTY_BILLING_ACCOUNT
ACTIVESHIPPING_UPS_THIRD_PARTY_BILLING_POSTAL_CODE
ACTIVESHIPPING_UPS_THIRD_PARTY_BILLING_COUNTRY_CODE


test/unit/carriers/ups_test.rb
```diff
@@ -332,6 +332,21 @@ def test_label_request_prepay
...
+                                           location_fixtures[:annapolis],
+                                           package_fixtures.values_at(:chocolate_stuff),
+                                           :test => true,
+                                           :bill_third_party => true,
+                                           :billing_account=>"A01B24",
+                                           :billing_zip=>"97013",
+                                           :billing_country => "US")
+
+    bill_third_party = response.search '/ShipmentConfirmRequest/Shipment/PaymentInformation/BillThirdParty'
+    refute_empty bill_third_party
```
trishume: I would also test that at least one of the options is being correctly passed through into the XML.

test/remote/ups_test.rb
```diff
@@ -218,6 +218,35 @@ def test_obtain_shipping_label_without_dimensions
...
+      }
+    )
+
+    assert response.success?
+
+    # All behavior specific to how a LabelResponse behaves in the
+    # context of UPS label data is a matter for unit tests.  If
+    # the data changes substantially, the create_shipment
+    # ought to raise an exception and this test will fail.
+    assert_instance_of ActiveShipping::LabelResponse, response
```
trishume: Does the response from UPS contain any data to suggest that it recognized the third party billing provider? Because currently none of the testing code can tell if the XML format passed to UPS is correct or if it just isn't throwing invalid schema errors. That would be a good thing to test, I know you have probably validated this but it is good to have it in the tests to catch if we break it or if UPS changes their API.

trishume:
Looks pretty good, just some testing concerns. Thanks for the contribution! :+1: 

Also if you have any pointers on how Shopify could get a set of these third party credentials for our internal testing we would be grateful.

test/remote/ups_test.rb
nrevko: When you create_shipment, for the most part, it just doesn't blow up. ShipmentAcceptResponse or ShipmentConfirmResponse do not contain any specific tag indicating that it is a third party billing. There is an indication on the gif label that the billing is "3rd party", but I am not sure how that can be parsed out from GraphicImage and be tested.

nrevko:
I've checked with the person who provided me with the 3rd party billing credentials. We have a partner we work with that created the credentials through their UPS representative. I do not know how this is done.

From the UPS XML documentation about account validation: 

> The account must be a valid UPS account number that is active. For US, PR and CA accounts, the account must be either a daily pickup account, an occasional account, or a customer B.I.N account, or a drop shipper account. All other accounts must be either a daily pickup account, an occasional account, a drop shipper account, or a non-shipping account.

There might be a flag that needs to be flipped on the UPS side, but that's my speculation.

test/remote/ups_test.rb
nrevko: ShipmentAcceptResponse and ShipmentConfirmResponse XML example from the above test: https://gist.github.com/nrevko/a9fdb4164a1272c732d8

trishume: Ok that's fine. Not great but nothing we can reasonably do. Fine as it is. :+1:

PR #290, opened 2015-06-12
Add maximum address field length to carriers

xuorig:
Certain carriers limit the address field to a certain amount of characters which cuts off the address coming from checkout when it is longer.

With this PR we'll be able to validate before that happens.

Also ```s/test:units/test:unit```

r: @trishume

### DIFF
--- a/README.md
+++ b/README.md
@@ -98,7 +98,7 @@ Active Shipping is currently being used and improved in a production environment
 
 ## Running the tests
 
-After installing dependencies with `bundle install`, you can run the unit tests with `rake test:units` and the remote tests with `rake test:remote`. The unit tests mock out requests and responses so that everything runs locally, while the remote tests actually hit the carrier servers. For the remote tests, you'll need valid test credentials for any carriers' tests you want to run. The credentials should go in ~/.active_shipping/credentials.yml, and the format of that file can be seen in the included [credentials.yml](https://github.com/Shopify/active_shipping/blob/master/test/credentials.yml). For some carriers, we have public credentials you can use for testing: see `.travis.yml`.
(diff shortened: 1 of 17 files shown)


trishume:
Looks good :+1: 

Other than me being a bit worried about the arbitrary `150` character general limit, it's small enough that it could conceivably break things. Especially for international carriers where us Canadians don't have as much knowledge, maybe they have super long addresses all on one line there. IMO it should be higher.

xuorig:
Yeah you're right, we will be safer with a higher limit, what would you say would be safe? ```500``` ?

xuorig:
```300``` seems like it is big enough to accommodate even huge addresses. Also added Api doc links to know where these numbers come from.

christianblais:
Just as an FYI, in Shopify, the hard limit is `255`.

trishume:
Great! If you want to use this soon I can put together a release this afternoon or Monday but otherwise I'll batch it up with some other stuff later.

I just checked and I used to have an address that was 149 characters if you put it on one line (which I did once when I didn't know what I was doing). Weird case though since university residences have crazy addresses.

kmcphillips:
FWIW I'm onboard with `255` because `VARCHAR`

kmcphillips:
:+1:

xuorig:
Alright made the limit 255, @trishume if its good with you, you can release whenever you're ready :)

christianblais:
:heart:

trishume:
You forgot a test, could you fix that and then I might be able to release this afternoon. :stuck_out_tongue_winking_eye:

christianblais:
> You forgot a test, could you fix that and then I might be able to release this afternoon. :stuck_out_tongue_winking_eye:

Go go go!

xuorig:
oops sorry :cry:

fixed!

trishume:
:neutral_face: Looks like I only read the file contents when I reviewed this PR and not the file names, and so did you. You added a bunch of your tests to the remote test files, easy mistake to make since fuzzy file finders return them first since the paths are shorter. When the remote tests are run (have to have credentials locally, which is why CI passed) shit fails because you didn't know to update them since they weren't failing for you.

Oh well, I'll fix them in master. ¯\\\_(ツ)\_/¯

xuorig:
aw damn -_-, thanks @trishume

PR #292, opened 2015-06-15
Fix tests and things up for a release.

trishume:
This fixes up the tests being in the wrong place from #290 and a date from #279. Also cleans up a test from #283.

Then it bumps the version number to `v1.3.0` according to semver because of the new features in this release.

This PR is mostly fixing currently broken tests in a trivial way so I'm not going to wait for review.

cc @kmcphillips


kmcphillips:
:+1:

PR #294, opened 2015-06-16
Add rake console command with handy things

trishume:
This adds a console for debugging ActiveShipping code and also more importantly the APIs of the carriers. 

In the past when I'm trying to figure out something about a carrier's API I have used an IRB session with a bunch of stuff loaded and had to copy-paste a bunch of things from tests and load a bajillion files.

This PR adds a `rake console` command based on pry that has a bunch of handy things pre-defined and included:
- ActiveShipping and testing fixtures and credentials modules included
- An equivalent to the `reload!` method from 'rails console` that reloads all of the active_shipping code
- A bunch of predefined carrier objects using the local test credentials from the remote tests
- A method for pretty-printing XML strings (unfortunately too commonly needed)

@kmcphillips @wvanbergen cc @RichardBlair because he could have used this in his weekend debugging session

### DIFF
--- a/test/console.rb
+++ b/test/console.rb
@@ -1,0 +1,40 @@
+require 'pry'
+require 'active_shipping'
+require 'test_helper'
+
+include ActiveShipping
+include ActiveShipping::Test::Credentials
+include ActiveShipping::Test::Fixtures
+
+def create_carrier(klass, creds)
+  options = credentials(creds).merge(:test => true)
+  klass.new(options)
+rescue NoCredentialsFound
+  STDERR.puts "No credentials found for #{creds}"
+  nil
+end
+
+def px(xml_s)
+  puts Nokogiri.XML(xml_s)
+end
+
+def reload!
+  # Supress a billion constant redefinition warnings
+  warn_level = $VERBOSE
+  $VERBOSE = nil
+
+  files = $LOADED_FEATURES.select { |feat| feat =~ /\/active_shipping\// }
+  files.each { |file| load file }
+
+  $VERBOSE = warn_level
+  files
+end
+
+# Prebuilt objects for most common carriers
+@canpo    = create_carrier(CanadaPost,:canada_post)
+@fedex    = create_carrier(FedEx,:fedex)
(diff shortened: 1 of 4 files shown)


kmcphillips:
Is this necessary? Is `pry --gem` not sufficient?

trishume:
The test helpers and credentials are really necessary to get anything done. This has already helped me significantly in looking into UPS surepost stuff. The pre-created carrier objects and reloading also save a lot of time. 

The reason Active Shipping needs this but other gems don't as much is all the API integrations require lots of valid frippery just to test them, and they need a lot of manual testing because the docs are often insufficient, this saves the effort of finding and including all that frippery.

https://www.youtube.com/watch?v=8To-6VIJZRE

kmcphillips:
You make a strong argument.

test/console.rb
```diff
@@ -0,0 +1,40 @@
...
+  files
+end
+
+# Prebuilt objects for most common carriers
+@canpo    = create_carrier(CanadaPost,:canada_post)
+@fedex    = create_carrier(FedEx,:fedex)
+@shipwire = create_carrier(Shipwire,:shipwire)
+@ups      = create_carrier(UPS,:ups)
+@usps     = create_carrier(USPS,:usps)
+# Tips: call reload! to reload all the active_shipping files, use fixtures from test_helpers for parameters
```
kmcphillips: You can `puts` these tips on console load.

kmcphillips:
:+1:

wvanbergen:
:+1: - a good development and test environment would really make working on this gem a lot easier. Maybe document it in the wiki/README/CONTRIBUTING?

test/console.rb
wvanbergen: You may also want to add some walrus facts to that list :troll:

trishume: I was going to `puts` them but the thing is that `binding.pry` prints the context so it gets printed anyways, the puts would be redundant. That's the reason that it is right above the binding.pry instead of next to the related methods.

PR #296, opened 2015-06-16
Add names for UPS SurePost services

trishume:
This should be all that is required to support USPS SurePost rate fetching on accounts that have a contract for it.

I can't test that this works without an account but I have some evidence from the interwebs:
- http://stackoverflow.com/questions/27231907/ups-surepost-api
- https://github.com/typefoo/node-shipping-ups/commit/08ecb7003867f3db915f267127cbfc2d80bd7a81
- https://github.com/pullingshots/Shipment/commit/5523534a6c43adef580fc78290d17af597ed49f4

@kmcphillips cc @louiskearns @shawnfrancis

### DIFF
--- a/lib/active_shipping/carriers/ups.rb
+++ b/lib/active_shipping/carriers/ups.rb
@@ -65,8 +65,11 @@ module ActiveShipping
       "83" => "UPS Today Dedicated Courier",
       "84" => "UPS Today Intercity",
       "85" => "UPS Today Express",
-      "86" => "UPS Today Express Saver"
-
+      "86" => "UPS Today Express Saver",
+      "92" => "UPS SurePost (USPS) < 1lb",
+      "93" => "UPS SurePost (USPS) > 1lb",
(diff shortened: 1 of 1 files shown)


lib/active_shipping/carriers/ups.rb
```diff
@@ -65,8 +65,11 @@ class UPS < Carrier
       "83" => "UPS Today Dedicated Courier",
       "84" => "UPS Today Intercity",
       "85" => "UPS Today Express",
-      "86" => "UPS Today Express Saver"
-
+      "86" => "UPS Today Express Saver",
+      "92" => "UPS SurePost (USPS) < 1lb",
+      "93" => "UPS SurePost (USPS) > 1lb",
+      "94" => "UPS SurePost (USPS) BPM",
+      "95" => "UPS SurePost (USPS) Media"
```
kmcphillips: Want to add the trailing `,` so the next diff doesn't have to touch this line? /nitpick

kmcphillips:
LGTM

PR #298, opened 2015-06-19
Add public remote test creds and fix test

trishume:
Adds back the creds removed in https://github.com/Shopify/active_shipping/pull/284 because I changed the password back to what it was.

Also adds an old Jaded Pixel (former name for Shopify Inc.) login for USPS. Was assured by USPS that nothing too bad could happen if it was released publicly.

Also fixes a test that failed only on Fridays :neutral_face:.

@kmcphillips @RichardBlair


jnormore:
> Also fixes a test that failed only on Fridays

lol

wvanbergen:
This is the best :+1: Thanks for putting all this work on making this project more healthy!

RichardBlair:
:heart:

PR #299, opened 2015-06-24
Fix bug with batch tracking errors

trishume:
An XPath that was designed without the idea of batched responses was giving back all the errors in a batch tracking request concatenated together instead of a specific error. This fixes the XPath and tests it.

@kmcphillips @garethson


garethson:
:boat:

PR #300, opened 2015-06-26
Fix bugs in USPS tracking

trishume:
While running batch processing on tracking numbers I discovered a few bugs in active_shipping USPS tracking. This PR fixes two of them:
- Some country names aren't the same as they are in ActiveUtils and have to be mapped separately. I only found 2 out of the 6000 numbers I tested but I have emailed someone at USPS asking if they have a full list.
- One type of event ("Origin Post is Preparing Shipment") doesn't have a time, even though it really should. This currently causes an exception. I handled this by returning an arbitrary date in the past (Jan 1, 2000).

I'm definitely open to discussion on how I handled the lack of date. I imagine lots of code would assume that all events have a time, which is true for everything except this rare event type, which is why I still return a time rather than nil. I chose an arbitrary past date rather than the current date so that the sort by time would do the right thing and put this event in the past.

@kmcphillips @garethson

### DIFF
--- a/lib/active_shipping/carriers/usps.rb
+++ b/lib/active_shipping/carriers/usps.rb
@@ -243,7 +243,14 @@ module ActiveShipping
         description = prefix
       end
 
-      timestamp = "#{node.at('EventDate').text}, #{node.at('EventTime').text}"
+      if node.at('EventDate').text.present?
+        timestamp = "#{node.at('EventDate').text}, #{node.at('EventTime').text}"
+        time = Time.parse(timestamp)
+      else
+        # Arbitrary time in past, because we need to sort and this is just dumb
+        time = Time.parse("Jan 01, 2000")
+      end
+
       event_code = node.at('EventCode').text
       city = node.at('EventCity').try(:text)
       state = node.at('EventState').try(:text)
@@ -255,7 +262,6 @@ module ActiveShipping
       # USPS returns upcased country names which ActiveUtils doesn't recognize without translation
       country = find_country_code_case_insensitive(country)
 
-      time = Time.parse(timestamp)
       zoneless_time = Time.utc(time.year, time.month, time.mday, time.hour, time.min, time.sec)
       location = Location.new(city: city, state: state, postal_code: zip_code, country: country)
       EventDetails.new(description, time, zoneless_time, location, event_code)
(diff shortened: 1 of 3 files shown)


lib/active_shipping/carriers/usps.rb
```diff
@@ -154,6 +154,11 @@ class USPS < Carrier
       "WS" => "Western Samoa"
     }
 
+    TRACKING_ODD_COUNTRY_NAMES = {
```
garethson: This is fine. The way we solved a similar problem in shopify is defining a `name_alternates` array in the countries YAML file (c[:name]), and then we roll that up into another key https://github.com/Shopify/shopify/blob/master/app/models/country_db.rb#L133). It's a bit more of a sustainable long term approach. Ok with this too though.

lib/active_shipping/carriers/usps.rb
```diff
@@ -238,7 +243,14 @@ def extract_event_details(node)
         description = prefix
       end
 
-      timestamp = "#{node.at('EventDate').text}, #{node.at('EventTime').text}"
+      if node.at('EventDate').text.present?
```
garethson: Dunno if this is our preferred style, but doing 
```ruby
time = if node.at('EventDate').text.present?
  x
else
  y
end
```

lib/active_shipping/carriers/usps.rb
```diff
@@ -238,7 +243,14 @@ def extract_event_details(node)
...
+        timestamp = "#{node.at('EventDate').text}, #{node.at('EventTime').text}"
+        time = Time.parse(timestamp)
+      else
+        # Arbitrary time in past, because we need to sort and this is just dumb
```
garethson: Could probably kill `and this is just dumb`. :scream:

garethson:
:sail

lib/active_shipping/carriers/usps.rb
trishume: There's lots of other hashes like this in the same place so I thought I would do it a consistent way, and there should only be a few of them.

trishume:
I received a list of the country names USPS uses from someone at USPS and reconciled the code so that it should handle any country they can throw at us.

kmcphillips:
This is a great approach. Huge. Thanks for digging to find that country list.

PR #302, opened 2015-06-30
Remove reference to undefined variable if using FedEx logger.

gblache:
That was causing an error when doing:
`ActiveShipping::FedEx.logger = Logger.new($stdout)`

### DIFF
--- a/lib/active_shipping/carriers/fedex.rb
+++ b/lib/active_shipping/carriers/fedex.rb
@@ -171,8 +171,6 @@ module ActiveShipping
       request = build_shipment_request(origin, destination, packages, options)
       logger.debug(request) if logger
 
-      logger.debug(confirm_response) if logger
-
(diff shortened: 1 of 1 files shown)


kmcphillips:
Why was this particular logging call causing an error? Why did you need to remove it?

gblache:
If you setup a logger for FedEx it will run the line that I removed in this PR.

That line references `confirm_response` which doesn't exist and throws an error.

kmcphillips:
Oh, I see now.

Thanks.

gblache:
Yeah sorry, I should have been more clear!  It was calling it in `create_shipment`.  Anyways, thanks!

kmcphillips:
Yep. Just needed a clearer description.

This will get lumped into our next release, but I'm not going to release just for this small fix. cc @trishume

PR #310, opened 2015-08-10
Use new USPS staging server, and don't use it

trishume:
USPS changed their testing server, at least for insecure tracking (not sure about the others).

This fixes the error `API Disabled: TrackV2.  This test request is no longer valid. Please use http://stg-production.shippingapis.com/ShippingAPI.dll` when running tests.

The thing is the new test server can't find any of the production tracking numbers that we use for testing and I can't find any tracking numbers that work on the server. So we now use the production server for the remote tracking tests, which since it is just tracking shouldn't be a big deal.

@kmcphillips


kmcphillips:
:+1: 

Give a version bump too.

trishume:
I generally bump the version on master in the same commit as I update the changelog right before I release, is that wrong?

kmcphillips:
Nope. totally fine

PR #316, opened 2015-09-15
added zpl support

RapsIn4:
- I added the ability to pass in an image_format option for UPS labels
- only have explicit support for zpl, but defaults to gif


lib/active_shipping/carriers/ups.rb
```diff
@@ -450,15 +450,23 @@ def build_shipment_request(origin, destination, packages, options={})
...
+              xml.LabelStockSize do
+                xml.Height('4')
+                xml.Width('6')
+              end
+            end
             xml.LabelPrintMethod do
-              xml.Code('GIF')
+               xml.Code(image_format)
             end
             xml.HTTPUserAgent('Mozilla/4.5') # hmmm
```
eboswort: According to the UPS docs this HTTPUserAgent node is actually only required if the LabelPrintMethod == 'GIF', and the LabelStockSize above actually applies to all of the non GIF formats (EPL, ZPL, STARPL and SPL), so would you mind flipping this conditional around so that the HTTPUserAgent is included if the LabelPrintMethod is GIF and the LabelStockSize is included in all other cases?

eboswort: Also maybe it'd be nice to just throw in a comment that lists EPL, ZPL, STARPL and SPL as the other options so people might be able to get away with not looking it up in the UPS docs.

eboswort:
Would you be able to write a quick spec to show that you get a successful response from UPS? You could model it off the other label creation specs and just add the 'ZPL' format and show that you got a good response back.

lib/active_shipping/carriers/ups.rb
eboswort: Sorry I don't mean to overwhelm you with comments but I actually just noticed that the UPS docs also say that 'LabelImageFormat' is only required when the LabelPrintMethod == 'GIF' so it might make sense to include that in the conditional so it was something like
```
if image_format == 'GIF'
  xml.HTTPUserAgent('Mozilla/4.5')
  xml.LabelImageFormat(image_format) do
  xml.Code(image_format)
else
  xml.LabelStockSize do
    xml.Height('4')
    xml.Width('6')
  end
end
```

RapsIn4:
I might add the response later.

eboswort:
@RapsIn4 Is this something you're really eager to merge? I think we could merge it without that test if you need to start using zebra format since it's so simple. But also if you wanna try just adding the following test and seeing if it works that would be great.

```
  def test_obtain_shipping_label_zpl_format
    response = @carrier.create_shipment(
      location_fixtures[:beverly_hills],
      location_fixtures[:new_york_with_name],
      package_fixtures.values_at(:american_wii),
      :image_format => 'ZPL',
      :test => true
    )

    assert response.success?
    assert_instance_of ActiveShipping::LabelResponse, response
  end
```

eboswort:
@RapsIn4 Hey, thanks for adding that test. Looks like the test failures are from something else.

I think you might have added another commit by accident though? The "added better international forms support" one.

That looks like it could be part of a different PR. If you could revert that commit then I think we could merge the ZPL stuff.

PR #319, opened 2015-09-23
Add compare_price to rate estimate

jnormore:
@kmcphillips @garethson for review please, thanks.


garethson:
Fine with this. Seeing as we have `insurance_price` I don't think it's unreasonable.

kmcphillips:
Yep. Good addition.

Can you include the version bump please?

PR #320, opened 2015-09-29
add phone required to rate estimate

aaslamin:
### Problem

Some shipping rates require a phone number.

### Solution

Add phone required to the shipping estimate.

Review: @kmcphillips @mdking 
cc: @jnormore

### DIFF
--- a/lib/active_shipping/rate_estimate.rb
+++ b/lib/active_shipping/rate_estimate.rb
@@ -55,6 +55,10 @@ module ActiveShipping
   #   The comparable price in cents
   #   @return [Integer]
   #
+  # @!attribute phone_required
+  #   Specifies if a phone number is required for the shipping rate.
+  #   @return [Boolean]
+  #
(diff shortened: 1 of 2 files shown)


lib/active_shipping/rate_estimate.rb
```diff
@@ -64,7 +64,7 @@ class RateEstimate
                 :shipping_date, :delivery_date, :delivery_range,
                 :currency, :negotiated_rate, :insurance_price,
                 :estimate_reference, :expires_at, :pickup_time,
-                :compare_price
+                :compare_price, :phone_required
```
mdking: Should you add this parameter to the above documentation?

kmcphillips:
:+1:

lib/active_shipping/rate_estimate.rb
aaslamin: Donezo

PR #324, opened 2015-10-23
Allow a failed response not to raise

mdking:
### Summary
Having this class raise on initialize if `success` is `false` is a huge pain. Allow an option to safely change the behaviour for users who do not want the raise behaviour while still supporting the existing behaviour.

Review @RichardBlair @lucasuyezu @jonathankwok

### DIFF
--- a/lib/active_shipping/response.rb
+++ b/lib/active_shipping/response.rb
@@ -18,12 +18,13 @@ module ActiveShipping #:nodoc:
     #   of a request executed against the sandbox or test environment of the carrier's API.
     # @option options [String] :xml The raw XML of the response.
     # @option options [String] :request The payload of the request.
+    # @option options [Boolean] :allow_failure Allows a failed response without raising.
     def initialize(success, message, params = {}, options = {})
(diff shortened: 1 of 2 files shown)


RichardBlair:
> Having this class raise on initialize if success is false is a huge pain

For documentation sake can you elaborate?

mdking:
There were some issues with this when Tristan was here dealing with the Fedex response issue. As well it easily leads developers down the path that they check for `success?` without realizing the class raises on failures and will never hit their logic. Of course testing will catch this in the end.

I think exceptions are great when used correctly but enforcing that a `Response` must be successful otherwise having to handle the raise is a little heavy handed imo.

lucasuyezu:
Can we have a regression test asserting that an exception is still raised without the flag?

lucasuyezu:
Ok, nevermind we already have that test.

RichardBlair:
Looks good. :shipit:

jonathankwok:
:ship:

mdking:
Rebased to recent master for CI, no changes.

PR #313, opened 2015-09-08
Fix Canada Post PWS remote tests

dangerdogz:
## Fix credentials for Canada Post PWS remote tests

Credentials taken from : [CPWS Rest fundamentals]
(https://www.canadapost.ca/cpo/mc/business/productsservices/developers/services/fundamentals.jsf)

Contract number is not needed for current ActiveShipping features, currently in the process of integrating contract_shipment support on my fork

CANADA_POST_PWS_API_KEY is first half of the API_KEY test value from above link and correspond to the "username"
CANADA_POST_PWS_SECRET is last half of the API_KEY test value from above link and correspond to the "password"

If you have an actual business account, you can get those value by logging in your account and accessing the "API keys" link

Values included in .travis.yml are actual test values provided by Canada Post

PR #325, opened 2015-11-10
Include event type codes for FedEx tracking events

jonathankwok:
### Problem
`ShipmentEvents` for FedEx weren't initialized with their `<EventType>`s. This results in no status code being associated with a `ShipmentEvent`.

### Solution
Save the `<EventType>` , much like we do with [USPS `ShipmentEvents`.](https://github.com/Shopify/active_shipping/blob/master/lib/active_shipping/carriers/usps.rb#L254)

Review: @mdking @lucasuyezu @cyprusad

### DIFF
--- a/lib/active_shipping/carriers/fedex.rb
+++ b/lib/active_shipping/carriers/fedex.rb
@@ -621,11 +621,12 @@ module ActiveShipping
 
           location = Location.new(:city => city, :state => state, :postal_code => zip_code, :country => country)
           description = event.at('EventDescription').text
+          type_code = event.at('EventType').text
 
           time          = Time.parse(event.at('Timestamp').text)
           zoneless_time = time.utc
 
-          shipment_events << ShipmentEvent.new(description, zoneless_time, location)
+          shipment_events << ShipmentEvent.new(description, zoneless_time, location, description, type_code)
         end
         shipment_events = shipment_events.sort_by(&:time)
 
--- a/test/remote/canada_post_pws_test.rb
+++ b/test/remote/canada_post_pws_test.rb
@@ -121,7 +121,7 @@ class RemoteCanadaPostPWSTest < Minitest::Test
   end
 
   def test_create_shipment
-    #skip "Failing with 'Contract Number is a required field' after API change, skipping because no clue how to fix, might need different creds"
+    skip "Failing with 'Contract Number is a required field' after API change, skipping because no clue how to fix, might need different creds"
     opts = {:customer_number => @customer_number, :service => "DOM.XP"}
     response = @cp.create_shipment(@home_params, @dom_params, @pkg1, @line_item1, opts)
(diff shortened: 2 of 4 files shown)


kmcphillips:
:+1: other than red CI

jonathankwok:
@kmcphillips Appears that all the fails are related to [this change?](https://github.com/Shopify/active_shipping/commit/d6d9b1bc6aad8e4c930b6a72bb519636c2175bbe)

My guess is there was some kind of an API change?

lib/active_shipping/carriers/fedex.rb
```diff
@@ -621,11 +621,12 @@ def parse_tracking_response(response, options)
 
           location = Location.new(:city => city, :state => state, :postal_code => zip_code, :country => country)
           description = event.at('EventDescription').text
+          type_code = event.at('EventType').text
 
           time          = Time.parse(event.at('Timestamp').text)
           zoneless_time = time.utc
 
-          shipment_events << ShipmentEvent.new(description, zoneless_time, location)
+          shipment_events << ShipmentEvent.new(description, zoneless_time, location, description, type_code)
```
lucasuyezu: Why are you passing `description` twice?

jonathankwok: It was that, or pass `nil`. USPS opted to [add `description` twice](https://github.com/Shopify/active_shipping/blob/master/lib/active_shipping/carriers/usps.rb#L626) so I went with that.

lucasuyezu: Oh I see. :+1:

cyprusad: Is that for `name` and `message` respectively? https://github.com/Shopify/active_shipping/blob/master/lib/active_shipping/shipment_event.rb#L5

jonathankwok: Yeah, it's always been using `description` for `name`, I just went ahead and passed `description` for `message` as well instead of passing in `nil`.

jonathankwok:
After some digging, I've found out that the customer number added in https://github.com/Shopify/active_shipping/pull/313 is for [contract shipping (see: _Test values for contract shipping, rating and returns_](https://www.canadapost.ca/cpo/mc/business/productsservices/developers/services/fundamentals.jsf). The customer number is for contract-shipping, which requires a contract number and a request built for contract-shipping.

Our `canada_post_pws.rb` uses the non-contracted shipment endpoint, so the credentials provided (and used in Travis) won't work for the failing tests. The only solution I have for now is to go back to the previous behaviour where we skipped those specific failing tests. We'll need different creds if we want those tests to pass - I have no idea how @dangerdogz was able to get them to pass on https://github.com/Shopify/active_shipping/pull/313.

kmcphillips:
Thanks for the digging @jonathankwok. I'm fine with `skip` since you have made a good case for it.

PR #326, opened 2015-11-13
Add UPS support for other labels like ZPL, EPL STARPL and SPL

wassimk:
This fixes and closes PR #316 too.

### DIFF
--- a/lib/active_shipping/carriers/ups.rb
+++ b/lib/active_shipping/carriers/ups.rb
@@ -481,15 +481,27 @@ module ActiveShipping
             end
           end
 
-          # I don't know all of the options that UPS supports for labels
-          # so I'm going with something very simple for now.
+          # Supported label formats:
+          # GIF, EPL, ZPL, STARPL and SPL
+          label_format = options[:label_format] ? options[:label_format].upcase : 'GIF'
+          label_size = options[:label_size] ? options[:label_size] : [4, 6]
+
           xml.LabelSpecification do
-            xml.LabelPrintMethod do
-              xml.Code('GIF')
+            xml.LabelStockSize do
+              xml.Height(label_size[0])
+              xml.Width(label_size[1])
             end
-            xml.HTTPUserAgent('Mozilla/4.5') # hmmm
-            xml.LabelImageFormat('GIF') do
-              xml.Code('GIF')
+
+            xml.LabelPrintMethod do
+              xml.Code(label_format)
+            end
+
+            # API requires these only if returning a GIF formated label
+            if label_format == 'GIF'
+              xml.HTTPUserAgent('Mozilla/4.5')
+              xml.LabelImageFormat(label_format) do
+                xml.Code(label_format)
+              end
             end
           end
         end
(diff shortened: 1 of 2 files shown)


lib/active_shipping/carriers/ups.rb
```diff
@@ -450,15 +450,26 @@ def build_shipment_request(origin, destination, packages, options={})
             end
           end
 
-          # I don't know all of the options that UPS supports for labels
-          # so I'm going with something very simple for now.
+          # Supported image_formats:
+          # GIF, EPL, ZPL, STARPL and SPL
+          label_format = options[:label_format].nil? ? 'GIF' : options[:label_format].upcase
```
kmcphillips: This still allows for `options[:label_format] = ""`

lib/active_shipping/carriers/ups.rb
```diff
@@ -450,15 +450,26 @@ def build_shipment_request(origin, destination, packages, options={})
...
+
             xml.LabelPrintMethod do
-              xml.Code('GIF')
+              xml.Code(label_format)
             end
-            xml.HTTPUserAgent('Mozilla/4.5') # hmmm
-            xml.LabelImageFormat('GIF') do
-              xml.Code('GIF')
+
+            xml.HTTPUserAgent('Mozilla/4.5')
```
kmcphillips: What's this about?

eboswort: In the UPS Api docs the description of this field is `Browser HTTPUserAgent String. This is the preferred way of identifying GIF image type to be generated`. They say it is `Required if ShipmentConfirmRequest/LabelSpecification/LabelPrintMethod/Code = GIF. Default to Mozilla/4.5 if this field is missing or has invalid value`

eboswort: Might be best to put this inside of the `if label_format == 'GIF'` block with a comment about how it's a stupid thing that UPS requires

lib/active_shipping/carriers/ups.rb
wassimk: I just reformatted it. If you don't specify a label format the API will return appropriate error. How does that look?

lib/active_shipping/carriers/ups.rb
kmcphillips: +1

lib/active_shipping/carriers/ups.rb
kmcphillips: :+1:

tmecklem:
I also need this support for epl image formats. I'll pitch in if anything else needs to be done to accept this PR.

kmcphillips:
Can you rebase, get the conflicts fixed, and the build green?

wassimk:
Hey @kmcphillips please let me know if we need to add anything else to merge this in.

kmcphillips:
Thanks for the good lookin' tests

kmcphillips:
Released in 1.6.3

tmecklem:
It looks like the height and width for labelstocksize might be backwards. I'm seeing height of 4 and width of 6 or 8, but the API docs say that valid values for width is 4 and height is 6 or 8.

tmecklem:
On trying out "correct" values for height and width, the api rejects the request. This seems to be a bug on UPS' side.

eboswort:
I've noticed the label images that come back from UPS are sideways. Maybe these two things are related.

PR #329, opened 2015-12-01
Ability to create a UPS return label

dholdren:
- Create a return label by setting the return_service_code
  - list of return service codes for reference
  - units used are now always based on shipper's country,
    allows international return from a different country than shipper
  - don't include SoldTo when it's a return
  - return requires a description at the package level

### DIFF
--- a/lib/active_shipping/carriers/ups.rb
+++ b/lib/active_shipping/carriers/ups.rb
@@ -93,6 +93,25 @@ module ActiveShipping
       "07" => "UPS Express"
     }
 
+    RETURN_SERVICE_CODES = {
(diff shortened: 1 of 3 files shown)


eboswort:
:+1: Looks like the failed build is unrelated

kmcphillips:
Yeah, it's the same USPS remote test.

kmcphillips:
:+1:

kmcphillips:
@dholdren I can ping you when I do a gem release. But for now you can use the github ref.

kmcphillips:
Released in 1.6.2

PR #328, opened 2015-12-01
Fedex "Home Delivery" is Tues-Sat

dholdren:
- use a different business days calculator
- option for absolute "pickup_date" instead of relative "turn_around_time"

PR #330, opened 2015-12-01
[WIP] Add support for both contract and non-contract for Canada Post

lucasuyezu:
Summary of changes:
- Removed constants throughout the code and changed them for small helper methods that know which value to return.
- `@options[:contract_id]` will tell which version of the payload must be generated.
- Added conditional base on `contract?` to have both implementations on the same file.

TODO:
Implement tests for both scenarios.

review @Shopify/shipping

### DIFF
--- a/lib/active_shipping/carriers/canada_post_pws.rb
+++ b/lib/active_shipping/carriers/canada_post_pws.rb
@@ -30,8 +30,6 @@ module ActiveShipping
 
     ENDPOINT = "https://soa-gw.canadapost.ca/"    # production
 
-    SHIPMENT_MIMETYPE = "application/vnd.cpc.ncshipment+xml"
-    RATE_MIMETYPE = "application/vnd.cpc.ship.rate+xml"
     TRACK_MIMETYPE = "application/vnd.cpc.track+xml"
     REGISTER_MIMETYPE = "application/vnd.cpc.registration+xml"
 
@@ -65,7 +63,7 @@ module ActiveShipping
     def find_rates(origin, destination, line_items = [], options = {}, package = nil, services = [])
       url = endpoint + "rs/ship/price"
       request  = build_rates_request(origin, destination, line_items, options, package, services)
-      response = ssl_post(url, request, headers(options, RATE_MIMETYPE, RATE_MIMETYPE))
+      response = ssl_post(url, request, headers(options, rate_mimetype, rate_mimetype))
       parse_rates_response(response, origin, destination)
     rescue ActiveUtils::ResponseError, ActiveShipping::ResponseError => e
       error_response(e.response.body, CPPWSRateResponse)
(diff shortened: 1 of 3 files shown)


lib/active_shipping/carriers/canada_post_pws.rb
```diff
@@ -564,6 +578,36 @@ def log(msg)
 
     private
 
+    def contract?
+      @options[:contract_id].present?
+    end
+
+    SHIPMENT_NON_CONTRACT_MIMETYPE = "application/vnd.cpc.ncshipment+xml"
```
mdking: The latest version for non-contract shipping is version 4, see https://www.canadapost.ca/cpo/mc/business/productsservices/developers/services/onestepshipping/createshipment.jsf

I can provide you with login credentials.

Otherwise :+1:  on the helper method.

test/remote/usps_test.rb
```diff
@@ -12,11 +12,11 @@ def setup
   end
 
   def test_tracking
-    response = @carrier.find_tracking_info('LN284529912US', test: false)
+    response = @carrier.find_tracking_info('LN757696446US', test: false)
```
lucasuyezu: @kmcphillips ^

kmcphillips: FYI @dholdren this fixes the tests we saw broken in #329 & #328.

mdking:
It is probably worth going through the API and ensuring the implementation reflects the latest versions for non-contract and contract both requests and responses. 

I believe the response to a the creation of a shipment for contract either contains new/update fields that the current implementation does not handle or are different from non-contract shipping.

See https://www.canadapost.ca/cpo/mc/business/productsservices/developers/services/shippingmanifest/createshipment.jsf with specific attention to the response `link` elements.

lucasuyezu:
We already have tests for the non-contract version. I'll write tests for the contract requests.

lib/active_shipping/carriers/canada_post_pws.rb
```diff
@@ -577,19 +621,27 @@ def tracking_url(pin)
 
     def create_shipment_url(options)
       raise MissingCustomerNumberError unless customer_number = options[:customer_number]
-      if @platform_id.present?
-        endpoint + "rs/#{customer_number}-#{@platform_id}/ncshipment"
+      if contract?
+        endpoint + "rs/#{customer_number}/#{customer_number}/shipment"
```
mdking: The second `customer_number` is actually the `mobo` (mail on behalf of). If specified it should be used or repeat the `customer_number` 

From the docs
>the mailed on behalf of customer number or repeat your customer number

lib/active_shipping/carriers/canada_post_pws.rb
```diff
@@ -564,6 +578,36 @@ def log(msg)
...
+
+    RATE_NON_CONTRACT_MIMETYPE = "application/vnd.cpc.ship.rate+xml"
+    RATE_CONTRACT_MIMETYPE = "application/vnd.cpc.ship.rate-v3+xml"
+    def rate_mimetype
+      contract? ? RATE_CONTRACT_MIMETYPE : RATE_NON_CONTRACT_MIMETYPE
+    end
+
+    RATE_CONTRACT_URL = "http://www.canadapost.ca/ws/ship/rate-v3"
+    RATE_NON_CONTRACT_URL = "http://www.canadapost.ca/ws/ship/rate"
+    def rate_url
```
garethson: I like this. Creating a variable just above the method, which describes what the difference is.

garethson:
:+1:

test/remote/usps_test.rb
kmcphillips: Pulled this into #331 and going to get it merged first.

kmcphillips: This is fixed on master. So you can rebase those changes in.

PR #333, opened 2015-12-04
Add ability to ship dry ice to UPS

ajsharma:
Adds the `DryIce` node to the purchase request when the packages have a `dry_ice` option.

Usage looks like:

```ruby
Package.new(80, [2, 6, 12], :units => :imperial, :dry_ice => {weight: 1.4})
```

and gets you a nice comment on the label

screen shot 2015-12-04 at 2 29 27 pm

and I'm hoping to use the same `dry_ice` hash option for FedEx soon.


kmcphillips:
> I'm hoping to use the same dry_ice hash option for FedEx soon.

Would be great to keep them consistent!

kmcphillips:
Thanks!

Issue #307, opened 2015-07-27
Upgrade to FedEx V18 rate WSDL - UK no service type issues


trishume:
According to FedEx the solution to the issues Shopify has been having with some requests within the UK receive rates but no service types is to upgrade to the "V18 rate WSDL". 

The reason is that the service types that would be returned in these requests aren't valid for the older version, so instead they make the response totally invalid by not including a crucial field :neutral_face: 

@kmcphillips 
cc @eboswort in case Casper has also noticed this issue

kmcphillips:
We aren't using the WSDL at all.

trishume:
The most recent docs on the FedEx site are for V16 :crying_cat_face:

eboswort:
Hmm, we actually haven't been shipping via Fedex lately at all, so I haven't run into any trouble. Looks like we're using version 13 of the API?

Do we just change this line: 
https://github.com/Shopify/active_shipping/blob/master/lib/active_shipping/carriers/fedex.rb#L290
to say 18 and then see what breaks?

kmcphillips:
It's possible that the XML API does support newer versions without the WSDL, and it's just not published anymore.

PR #334, opened 2015-12-08
Catch NoMethodErrors in FEDEX find_and_parse_rate method

heinzf:
See https://github.com/Shopify/active_shipping/issues/307

The ```parse_rate_response``` method would crash and return a NoMethodError whenever the service type was missing in the XML response from FEDEX.

This seems to be related to the new version of the upstream API. As a workaround before we implement this new version, this PR proposes to catch all NoMethodErrors in ```the parse_rate_response``` loop through the rates, and return as much as possible.

Because of this, the rates list might contain nil values, hence it will be compacted and also kept as successful, as not to raise an exception afterward, while logging the faulty XML response.

### DIFF
--- a/lib/active_shipping/carriers/fedex.rb
+++ b/lib/active_shipping/carriers/fedex.rb
@@ -474,31 +474,46 @@ module ActiveShipping
       message = response_message(xml)
 
       if success
+        missing_xml_field = false
         rate_estimates = xml.root.css('> RateReplyDetails').map do |rated_shipment|
-          service_code = rated_shipment.at('ServiceType').text
-          is_saturday_delivery = rated_shipment.at('AppliedOptions').try(:text) == 'SATURDAY_DELIVERY'
-          service_type = is_saturday_delivery ? "#{service_code}_SATURDAY_DELIVERY" : service_code
+          begin
+            service_code = rated_shipment.at('ServiceType').text
+            is_saturday_delivery = rated_shipment.at('AppliedOptions').try(:text) == 'SATURDAY_DELIVERY'
+            service_type = is_saturday_delivery ? "#{service_code}_SATURDAY_DELIVERY" : service_code
 
-          transit_time = rated_shipment.at('TransitTime').text if ["FEDEX_GROUND", "GROUND_HOME_DELIVERY"].include?(service_code)
-          max_transit_time = rated_shipment.at('MaximumTransitTime').try(:text) if service_code == "FEDEX_GROUND"
+            transit_time = rated_shipment.at('TransitTime').text if ["FEDEX_GROUND", "GROUND_HOME_DELIVERY"].include?(service_code)
+            max_transit_time = rated_shipment.at('MaximumTransitTime').try(:text) if service_code == "FEDEX_GROUND"
 
-          delivery_timestamp = rated_shipment.at('DeliveryTimestamp').try(:text)
-          delivery_range = delivery_range_from(transit_time, max_transit_time, delivery_timestamp, (service_code == "GROUND_HOME_DELIVERY"), options)
+            delivery_timestamp = rated_shipment.at('DeliveryTimestamp').try(:text)
+            delivery_range = delivery_range_from(transit_time, max_transit_time, delivery_timestamp, (service_code == "GROUND_HOME_DELIVERY"), options)
 
-          currency = rated_shipment.at('RatedShipmentDetails/ShipmentRateDetail/TotalNetCharge/Currency').text
-          RateEstimate.new(origin, destination, @@name,
-               self.class.service_name_for_code(service_type),
-               :service_code => service_code,
-               :total_price => rated_shipment.at('RatedShipmentDetails/ShipmentRateDetail/TotalNetCharge/Amount').text.to_f,
-               :currency => currency,
-               :packages => packages,
-               :delivery_range => delivery_range)
+            currency = rated_shipment.at('RatedShipmentDetails/ShipmentRateDetail/TotalNetCharge/Currency').text
+
+            RateEstimate.new(origin, destination, @@name,
+                 self.class.service_name_for_code(service_type),
+                 :service_code => service_code,
+                 :total_price => rated_shipment.at('RatedShipmentDetails/ShipmentRateDetail/TotalNetCharge/Amount').text.to_f,
+                 :currency => currency,
+                 :packages => packages,
+                 :delivery_range => delivery_range)
+          rescue NoMethodError
+            missing_xml_field = true
+            nil
+          end
         end
+
+        rate_estimates = rate_estimates.compact
+        logger.warn("[FedexParseRateError] Some fields where missing in the response: #{response}") if logger && missing_xml_field
 
         if rate_estimates.empty?
           success = false
-          message = "No shipping rates could be found for the destination address" if message.blank?
+          if missing_xml_field
+            message = "The response from the carrier contained errors and could not be treated"
+          else
+            message = "No shipping rates could be found for the destination address" if message.blank?
+          end
         end
+
       else
         rate_estimates = []
       end
--- a/test/unit/carriers/fedex_test.rb
+++ b/test/unit/carriers/fedex_test.rb
@@ -205,6 +205,32 @@ class FedExTest < Minitest::Test
     assert_equal message, exception.message
   end
 
+  def test_parsing_response_with_no_system_code_for_one_rate_does_not_crash
+    mock_response = xml_fixture('fedex/freight_rate_response').gsub('<v13:ServiceType>FEDEX_FREIGHT_ECONOMY</v13:ServiceType>','')
+
+    @carrier.expects(:commit).returns(mock_response)
+    @carrier.logger = Logger.new(StringIO.new)
+    @carrier.logger.expects(:warn).once.with("[FedexParseRateError] Some fields where missing in the response: #{mock_response}")
+
+    rates = @carrier.find_rates( location_fixtures[:ottawa], location_fixtures[:beverly_hills], package_fixtures.values_at(:book, :wii), :test => true)
+    assert rates.rates.length == 1
+    assert_equal rates.rates[0].service_code, 'FEDEX_FREIGHT_PRIORITY'
+  end
+
+  def test_parsing_response_with_no_system_code_on_any_shipping_rate
+    mock_response = xml_fixture('fedex/freight_rate_response').gsub('<v13:ServiceType>FEDEX_FREIGHT_ECONOMY</v13:ServiceType>','').gsub('<v13:ServiceType>FEDEX_FREIGHT_PRIORITY</v13:ServiceType>','')
+
+    @carrier.expects(:commit).returns(mock_response)
+    @carrier.logger = Logger.new(StringIO.new)
+    @carrier.logger.expects(:warn).once.with("[FedexParseRateError] Some fields where missing in the response: #{mock_response}")
+
+    exception = assert_raises(ActiveShipping::ResponseError) do
+      @carrier.find_rates( location_fixtures[:ottawa], location_fixtures[:beverly_hills], package_fixtures.values_at(:book, :wii), :test => true)
+    end
+    message = "The response from the carrier contained errors and could not be treated"
+    assert_equal exception.message, message
+  end
+
   def test_service_name_for_code
     FedEx::SERVICE_TYPES.each do |capitalized_name, readable_name|
       assert_equal readable_name, FedEx.service_name_for_code(capitalized_name)


lib/active_shipping/carriers/fedex.rb
```diff
@@ -475,7 +475,8 @@ def parse_rate_response(origin, destination, packages, response, options)
 
       if success
         rate_estimates = xml.root.css('> RateReplyDetails').map do |rated_shipment|
-          service_code = rated_shipment.at('ServiceType').text
+          service_code = rated_shipment.at('ServiceType').try(:text)
+          raise ActiveShipping::ResponseError, "The response did not contain service type" unless service_code.present?
```
christianblais: Since it is a change in their API, I'm afraid this will just continue to the next attribute that isn't gonna be define neither. Do we have a copy of the response we're receiving? Could we run a test against it?

heinzf: Ah I see what you mean now. I'll try to find it to see if there's something else broken you're right

mdking: As mentioned in slack I would offer up the suggestion of rather than directly raising the `ActiveShipping::ResponseError`, create a `RateResponse` with `success = false` similar to how it is performed for no rates here https://github.com/heinzf/active_shipping/blob/master/lib/active_shipping/carriers/fedex.rb#L500

This way the `ActiveShipping::ResponseError` will have the failed `RateResponse` with the xml and request. Would just want to verify that those fields are being logged somewhere.

heinzf:
@christianblais @mdking 
I changed things a bit following your comments.
I will try to "catch" missing service_type fields and log the xml. The rates list will then be compacted and kept as successful, as not to raise an exception afterward.

The idea is to log the bad behaving XML in a warning, but still give as much information as possible, as a best effort.

Tell me if the idea and the flow seem legit.

lib/active_shipping/carriers/fedex.rb
```diff
@@ -495,15 +500,23 @@ def parse_rate_response(origin, destination, packages, response, options)
                :delivery_range => delivery_range)
         end
 
-        if rate_estimates.empty?
+        if missing_xml_field && logger
+          logger.warn(xml)
+        end
+
```
mdking: I would `compact` once here on the `rate_estimates` it doesn't have to be repeated.

lib/active_shipping/carriers/fedex.rb
```diff
@@ -495,15 +500,23 @@ def parse_rate_response(origin, destination, packages, response, options)
...
+        if rate_estimates.compact.empty? && missing_xml_field
```
mdking: This can be rewritten:

```
if rates_estmates.empty?
  success = false
  
  if miss_xml_field
    ...
  else 
    ...
  end
end

File: test/unit/carriers/fedex_test.rb [hunk 4]
```diff
@@ -205,6 +205,32 @@ def test_parsing_response_with_no_rate_reply
...
+    @carrier.logger = Logger.new(StringIO.new)
+
+    exception = assert_raises(ActiveShipping::ResponseError) do
+      @carrier.find_rates( location_fixtures[:ottawa],
+                           location_fixtures[:beverly_hills],
+                           package_fixtures.values_at(:book, :wii), :test => true)
+    end
+    message = "The response from the carrier contained errors and could not be treated"
+    assert_equal exception.message, message
+  end
```
mdking: Can you add the log assert as part of this test?

heinzf: I've been trying for a while with no success yet :(

lib/active_shipping/carriers/fedex.rb
```diff
@@ -472,10 +472,15 @@ def parse_rate_response(origin, destination, packages, response, options)
...
+      missing_xml_field = false
 
       if success
         rate_estimates = xml.root.css('> RateReplyDetails').map do |rated_shipment|
-          service_code = rated_shipment.at('ServiceType').text
+          service_code = rated_shipment.at('ServiceType').try(:text)
+          unless service_code.present?
+            missing_xml_field = true
+            next
+          end
```
christianblais: Since this is a change in the API, I see that other errors could happen later on in the process, such as `rated_shipment.at('RatedShipmentDetails/ShipmentRateDetail/TotalNetCharge/Currency').text`. Since we probably don't want to wrap all the calls with a `if` like the above, should we simply rescue `NoMethodError`, in which we could log?

heinzf: @christianblais Indeed.
I've added a rescue/begin and kept the same "return what we can" logic

test/unit/carriers/fedex_test.rb
```diff
@@ -205,6 +205,31 @@ def test_parsing_response_with_no_rate_reply
     assert_equal message, exception.message
   end
 
+  def test_parsing_response_with_no_system_code_for_one_rate_does_not_crash
+    mock_response = xml_fixture('fedex/freight_rate_response').gsub('<v13:ServiceType>FEDEX_FREIGHT_ECONOMY</v13:ServiceType>','')
+
+    @carrier.expects(:commit).returns(mock_response)
+    @carrier.logger = Logger.new(StringIO.new)
+    @carrier.logger.expects(:warn)
```
mdking: Thanks for adding that, if you would like to be more explicit you can add:
`@carrier.logger.expects(:warn).with(mock_response).once`

lib/active_shipping/carriers/fedex.rb
```diff
@@ -474,31 +474,46 @@ def parse_rate_response(origin, destination, packages, response, options)
...
+            RateEstimate.new(origin, destination, @@name,
+                 self.class.service_name_for_code(service_type),
+                 :service_code => service_code,
+                 :total_price => rated_shipment.at('RatedShipmentDetails/ShipmentRateDetail/TotalNetCharge/Amount').text.to_f,
+                 :currency => currency,
+                 :packages => packages,
+                 :delivery_range => delivery_range)
+          rescue NoMethodError
+            missing_xml_field = true
+            next
```
christianblais: Is this required?

lib/active_shipping/carriers/fedex.rb
```diff
@@ -474,31 +474,46 @@ def parse_rate_response(origin, destination, packages, response, options)
...
+          end
         end
 
+        rate_estimates = rate_estimates.compact
+        logger.warn(xml) if logger && missing_xml_field
```
christianblais: I believe simply outputing `xml` might be confusing. Should we at least indicate there has been an error of some sort?

lib/active_shipping/carriers/fedex.rb
heinzf: The ```next``` statement? Without it it the iteration returns the assignment to missing_xml_field, and the ```compact``` method afterward cannot do its job.
That being said, ```next``` is overkill as we are already at the end of the block and we'll start the next iteration anyway, a simple nil will do the trick.

heinzf:
@christianblais @mdking 
I've added a log message and replace the if by a larger begin/rescue

christianblais:
Can you please provide a good PR description and meaningful commits please?

mdking:
Code LGTM. 

On our end when exception occurs for the `NoMethodError` it will be rescued and the response is logged. To log the warning our code will have to be modified to initialize the logger object when the carrier adapter is created.

lib/active_shipping/carriers/fedex.rb
```diff
@@ -474,31 +474,46 @@ def parse_rate_response(origin, destination, packages, response, options)
...
+                 :currency => currency,
+                 :packages => packages,
+                 :delivery_range => delivery_range)
+          rescue NoMethodError
+            missing_xml_field = true
+            nil
+          end
         end
 
+        rate_estimates = rate_estimates.compact
```
kmcphillips: What is this for? This is a behaviour change.

kmcphillips:
One comment. But mostly a big improvement. Thank you!

lib/active_shipping/carriers/fedex.rb
heinzf: I've updated the PR description to explain, it was in a comment bellow it :)

The method would loop through the XML elements for the rates with a ```array.map```.
If one of the iterations went through the ```begin/rescue```, the element would take the last value returned by the ```rescue```, which seems to be ```true```.
I've changed it to return ```nil``` so that we can compact the list and only keep the rates that were correctly parsed.

kmcphillips: Thanks for clarifying.

heinzf:
Is it :rocket: ready?

heinzf:
@mdking @kmcphillips @christianblais 
Sorry, still not used to the correct github ping policy :(

test/unit/carriers/fedex_test.rb
```diff
@@ -205,6 +205,31 @@ def test_parsing_response_with_no_rate_reply
     assert_equal message, exception.message
   end
 
+  def test_parsing_response_with_no_system_code_for_one_rate_does_not_crash
+    mock_response = xml_fixture('fedex/freight_rate_response').gsub('<v13:ServiceType>FEDEX_FREIGHT_ECONOMY</v13:ServiceType>','')
+
+    @carrier.expects(:commit).returns(mock_response)
+    @carrier.logger = Logger.new(StringIO.new)
+    @carrier.logger.expects(:warn).once
```
christianblais: Should expect with what. `expects(:warn).once.with(BLA BLA)`

test/unit/carriers/fedex_test.rb
```diff
@@ -205,6 +205,31 @@ def test_parsing_response_with_no_rate_reply
...
+    rates = @carrier.find_rates( location_fixtures[:ottawa], location_fixtures[:beverly_hills], package_fixtures.values_at(:book, :wii), :test => true)
+    assert rates.rates.length == 1
+    assert_equal rates.rates[0].service_code, 'FEDEX_FREIGHT_PRIORITY'
+  end
+
+  def test_parsing_response_with_no_system_code
+    mock_response = xml_fixture('fedex/freight_rate_response').gsub('<v13:ServiceType>FEDEX_FREIGHT_ECONOMY</v13:ServiceType>','').gsub('<v13:ServiceType>FEDEX_FREIGHT_PRIORITY</v13:ServiceType>','')
```
christianblais: Why twice?

test/unit/carriers/fedex_test.rb
```diff
@@ -205,6 +205,31 @@ def test_parsing_response_with_no_rate_reply
...
+
+    @carrier.expects(:commit).returns(mock_response)
+    @carrier.logger = Logger.new(StringIO.new)
+    @carrier.logger.expects(:warn).once
+    exception = assert_raises(ActiveShipping::ResponseError) do
+      @carrier.find_rates( location_fixtures[:ottawa], location_fixtures[:beverly_hills], package_fixtures.values_at(:book, :wii), :test => true)
+
```
christianblais: :scissors:

test/unit/carriers/fedex_test.rb
```diff
@@ -205,6 +205,31 @@ def test_parsing_response_with_no_rate_reply
...
+    assert rates.rates.length == 1
+    assert_equal rates.rates[0].service_code, 'FEDEX_FREIGHT_PRIORITY'
+  end
+
+  def test_parsing_response_with_no_system_code
+    mock_response = xml_fixture('fedex/freight_rate_response').gsub('<v13:ServiceType>FEDEX_FREIGHT_ECONOMY</v13:ServiceType>','').gsub('<v13:ServiceType>FEDEX_FREIGHT_PRIORITY</v13:ServiceType>','')
+
+    @carrier.expects(:commit).returns(mock_response)
+    @carrier.logger = Logger.new(StringIO.new)
+    @carrier.logger.expects(:warn).once
```
christianblais: Let's be explicit about the `with`.

test/unit/carriers/fedex_test.rb
```diff
@@ -205,6 +205,31 @@ def test_parsing_response_with_no_rate_reply
...
+    rates = @carrier.find_rates( location_fixtures[:ottawa], location_fixtures[:beverly_hills], package_fixtures.values_at(:book, :wii), :test => true)
+    assert rates.rates.length == 1
+    assert_equal rates.rates[0].service_code, 'FEDEX_FREIGHT_PRIORITY'
+  end
+
+  def test_parsing_response_with_no_system_code
```
christianblais: What does `system_code` mean?

test/unit/carriers/fedex_test.rb
heinzf: Damn, thanks.

heinzf: Ah no sorry, I thought I had copy pasted one of the tests somewhere. There's one test per behaviour, whether there's at least one rate found, or none. (Returns something vs returns an error)

christianblais: I was talking about the empty line.

heinzf:
@christianblais I've changed the title of the text to better explain that the system code belong to the shipping rate, made the empty lines cleaner, and added explicit validation for the content of the XML.

christianblais:
:+1:

kmcphillips:
Thanks @heinzf.

kmcphillips:
Released in 1.6.4

PR #336, opened 2015-12-11
Swap incorrect height and width defaults for UPS labels

wassimk:
In #326 it was noted the labels are coming in sideways possibly because the height and width are swapped. I tested it and they come through sideways no matter what. But, this corrects that typo regardless.


kmcphillips:
Can you get the tests passing please? Then I will merge.

kmcphillips:
Bump. If you want this released in the next version.

wassimk:
I did more testing and I believe the height and width are indeed swapped in the UPS API. I reverted my changes and documented the code more. I'm still doing testing though. I'm going to request a 4 x 8 label and see that it returns what's expected.

I think the current failed test for delivery estimate is because UPS doesn't deliver on New Years Day.

kmcphillips:
I'm going to close this out. If you get test covering it and some supporting documentation for this change, feel free to re-open.

PR #338, opened 2016-01-06
Added support for USPS attempted delivery dates

cwise:
We had a requirement to determine the first delivery attempt on a shipment as opposed to the actual delivery date.

For USPS, we pull the first event which is a failed delivery attempt.


kmcphillips:
:+1: Thanks

kmcphillips:
Released in 1.6.5

PR #339, opened 2016-01-12
Added support for multiple delivery attempts for UPS tracking

cwise:
Added some further enhancements on failed delivery attempts with support for UPS (to complement previous changes for USPS)


eboswort:
This is a nice feature! I'm a little worried about that test though. Maybe worth just testing with unit tests where you can mock the returns from ups and make sure it's parsed correctly?
I think putting in a test that we know is likely to start failing in some number of months might be a bad move.

jonathankwok:
Hey, @cwise, just took a look at your changes - I think it's great that you added support for attempted delivery attempts. That said, I'm not as on board with modifying the existing `TrackingResponse` class to have multiple `delivery_attempt` attributes. I think just one was enough, especially since other carriers only ever try to deliver once. 

That said, I think recording the first delivery attempt for UPS is still worth doing.

What are your thoughts, @kmcphillips @mdking ?

mdking:
I agree with @jonathankwok that I would like to avoid adding the multiple `delivery_attempt` attributes. It is a larger change in the code but what is everyone's thoughts on adding a `attempted_delivery_dates` attribute which is the array of `delivery_attempt` dates? 

There would be a bit of work to make sure it's backwards compatible with the existing `attempted_delivery_date` attribute.

kmcphillips:
@cwise I'm going to close this out for now. If you rebase, get test covering the change, get the build green, and get supporting docs. Feel free to re-open this PR.

Issue #341, opened 2016-01-19
UPS's build_delivery_dates_request fails if a package's value is nil


ajsharma:
Results in a TypeError when I don't provide a packages' value

```ruby
> get_delivery_date_estimates(origin, destination, packages)
TypeError: nil can't be coerced into Fixnum
from /active_shipping/lib/active_shipping/carriers/ups.rb:496:in `+'
```

Looks like the cause is here https://github.com/Shopify/active_shipping/blob/master/lib/active_shipping/carriers/ups.rb#L537

and `package.value` could be `package.value.to_i` to solve it, and/or we could avoid the `InvoiceLineTotal` block entirely unless at least one package has a value.

PR #342, opened 2016-01-19
Avoid exception during delivery date request

ajsharma:
Which occurs when one or more packages' value is nil.

Should fix https://github.com/Shopify/active_shipping/issues/341

If no packages have a value, then the entire block is
avoided to avoid an accidental 0 value. According to the
docs, `InvoiceLineTotal` is only required for "International
non-document shipments."

### DIFF
--- a/lib/active_shipping/carriers/ups.rb
+++ b/lib/active_shipping/carriers/ups.rb
@@ -532,10 +532,12 @@ module ActiveShipping
             xml.Weight([value.round(3), 0.1].max)
           end
 
-          xml.InvoiceLineTotal do
-            xml.CurrencyCode('USD')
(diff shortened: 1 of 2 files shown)


eboswort:
Looks good to me!

kmcphillips:
There is a subtle difference here. You explicitly want to omit the node if the value is `0`? Because otherwise you could solve the bug with a `|| 0`.

ajsharma:
@kmcphillips I'm afraid I don't understand your question. 

I want to omit the node only if _all_ of the packages have a `value` of `nil`. If any of the packages have a `0` (or any number) as their `value`, then the node would be included.

kmcphillips:
Ok cook, just making sure that was as intended. :+1:

PR #343, opened 2016-01-25
Add Saturday Delivery option to UPS rate call

KBalderson:


KBalderson:
looking at travis-ci, there is a problem with the SSL connection to the API that I cannot reproduce locally.

KBalderson:
Just found [this document](https://www.ups.com/media/en/value-added_pricing_daily.pdf), which indicates that, for services for which saturday delivery is an option, the cost is always $16. Feel free to close this pull request without merging unless you feel it adds some sort of value.

fiestacasey:
@KBalderson I suspect that issue is effecting master currently as well.  UPS is forcing customers to switch to using TLS 1.2 as per [this announcement](https://www.ups.com/content/us/en/resources/techsupport/data-security.html)

PR #345, opened 2016-02-11
Document FedEx login and test keys

adacosta:
I spent too much of my life figuring out that login keys were not the same between active_shipping and FedEx. I assume I'm not the first to experience this problem. Hopefully this documentation note will help others.


kmcphillips:
Thanks.

CI fail is a known SSL problem.

PR #346, opened 2016-02-16
Handle FedEx tracking response failure code 9045

adacosta:
I'm not sure of the source of this failure. I accidentally left a 6 second interval polling task open in a terminal for most of the day to the FedEx test endpoint, so maybe FedEx cut me off. It's also possible their test service is not functioning correctly at the moment. At any rate, it's better to handle the code because the error message and class which were returned to me weren't helpful.

If we don't handle this error, we get:
```
ActiveShipping::Error: Tracking response does not contain status code
```

If we do handle this error, we get:
```
ActiveShipping::ResponseContentError: Sorry, we are unable to process your tracking request.  Please retry later, or contact Customer Service at 1.800.Go.FedEx(R) 800.463.3339.
```

I dumped the response xml from a pry session to use as a test fixture and red/greened the fix. Here is the red output before it went green.

```
2) Failure:
FedExTest#test_response_failure_code_9045 [/Users/adacosta/projects/forks/active_shipping/test/unit/carriers/fedex_test.rb:334]:
[ActiveShipping::ResponseContentError] exception expected, not
Class: <ActiveShipping::Error>
Message: <"Tracking response does not contain status code">
---Backtrace---
/Users/adacosta/projects/forks/active_shipping/lib/active_shipping/carriers/fedex.rb:618:in `parse_tracking_response'
/Users/adacosta/projects/forks/active_shipping/lib/active_shipping/carriers/fedex.rb:161:in `find_tracking_info'
/Users/adacosta/projects/forks/active_shipping/test/unit/carriers/fedex_test.rb:335:in `block in test_response_failure_code_9045'
---------------
```

### DIFF
--- a/lib/active_shipping/carriers/fedex.rb
+++ b/lib/active_shipping/carriers/fedex.rb
@@ -605,6 +605,11 @@ module ActiveShipping
           else
             raise ActiveShipping::ResponseContentError, StandardError.new(first_notification.at('Message').text)
           end
+        elsif first_notification.at('Severity').text == 'FAILURE'
+          case first_notification.at('Code').text
+          when '9045'
+            raise ActiveShipping::ResponseContentError, StandardError.new(first_notification.at('Message').text)
+          end
         end
 
         tracking_number = tracking_details.at('TrackingNumber').text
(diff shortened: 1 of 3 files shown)


kmcphillips:
Thanks.

Test fails are the UPS remotes.

PR #350, opened 2016-03-18
Force TLS 1.2 for UPS

kmcphillips:
@garethson @jonathankwok @thegedge 

Get this damn build :green_heart: again.

We got this warning a while back:
security_upgrade_required_for_ups_developer_kit_apis_-_kevin_mcphillips_shopify_com_-_shopify_mail

Our servers and environment default to the proper TLS version, but TravisCI is not apparently. So we are forcing `:TLSv1_2` and that gets our build passing again.

### DIFF
--- a/lib/active_shipping/carriers/ups.rb
+++ b/lib/active_shipping/carriers/ups.rb
@@ -3,6 +3,7 @@
 module ActiveShipping
   class UPS < Carrier
     self.retry_safe = true
+    self.ssl_version = :TLSv1_2
 
     cattr_accessor :default_options
     cattr_reader :name


RichardBlair:
Checks out. :shipit:

kmcphillips:
The exception we were seeing, once I removed the wrapper, was:

```
OpenSSL::SSL::SSLError: SSL_connect returned=1 errno=0 state=unknown state: tlsv1 alert protocol version
```

garethson:
:tada:

thegedge:
Nice! Cool that they took this into consideration and started off with just the test environment early so people could catch it before the same change hits production. The Right Way™.

kmcphillips:
Thanks friends! :green_heart:

PR #351, opened 2016-03-21
Refactored setters on rate estimate

MalazAlamir:
Small refactor on `rate_estimate` that moves the private method to a protected visibility and creates a setter for `delivery_range`.

cc/ @jonathankwok @cyprusad

### DIFF
--- a/lib/active_shipping/rate_estimate.rb
+++ b/lib/active_shipping/rate_estimate.rb
@@ -63,7 +63,7 @@ module ActiveShipping
   #   The price of insurance in cents.
   #   @return [Integer]
   class RateEstimate
-    attr_reader :origin, :destination, :package_rates,
+    attr_accessor :origin, :destination, :package_rates,
                 :carrier, :service_name, :service_code,
                 :shipping_date, :delivery_date, :delivery_range,
                 :currency, :negotiated_rate, :insurance_price,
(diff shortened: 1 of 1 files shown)


jonathankwok:
LGTM. CC: @RichardBlair @kmcphillips

kmcphillips:
This changes the visibility of `date_for` too.

kmcphillips:
Is the setter going to be used elsewhere? Otherwise it should probably stay in the initializer. It breaks the existing pattern of making all the properties immutable.

RichardBlair:
@kmcphillips We had a discussion in a Shopify repo.

Essentially there are cases where you may want to initialize a `RateEstimate` and then fill in more details later. If you want to do this, you currently would need to write a setter, and copy the logic from the initializer in your setter. By moving this to a protected setter we encapsulate the logic, and allow subclasses of `RateEstimate` access those setters, and potentially expose them publicly when needed.

@MalazAlamir :shipit:

RichardBlair:
Actually, Malaz, can you create setters for all the attributes that are set to fill out the interface?

MalazAlamir:
@RichardBlair yep, will do. I'll create setters for the attributes that go through some sort of logic in `initialize`

MalazAlamir:
reping @jonathankwok @cyprusad after new changes

jonathankwok:
Nice, I like the reworked interface. :shipit:

cyprusad:
:ship:

PR #352, opened 2016-03-23
Change the returned arbitrary date to epoch

MalazAlamir:
Noticed that if we don't get an EventTime in the response we return arbitrary date in the past (1/1/2000). This is a little bit confusing, it would be clearer to return something that would clearly signal that the EventTime is missing, so I'm going with Time.at(0).

Thoughts?

cc/ @kmcphillips @jonathankwok @cyprusad

### DIFF
--- a/lib/active_shipping/carriers/usps.rb
+++ b/lib/active_shipping/carriers/usps.rb
@@ -247,8 +247,8 @@ module ActiveShipping
         timestamp = "#{node.at('EventDate').text}, #{node.at('EventTime').text}"
         Time.parse(timestamp)
       else
-        # Arbitrary time in past, because we need to sort properly by time
-        Time.parse("Jan 01, 2000")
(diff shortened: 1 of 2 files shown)


lib/active_shipping/carriers/usps.rb
```diff
@@ -247,8 +247,8 @@ def extract_event_details(node)
         timestamp = "#{node.at('EventDate').text}, #{node.at('EventTime').text}"
         Time.parse(timestamp)
       else
-        # Arbitrary time in past, because we need to sort properly by time
-        Time.parse("Jan 01, 2000")
+        # Epoch time, because we need to sort properly by time
+        Time.at(0)
```
kmcphillips: This is the Unix epoch? I think that's reasonable.

kmcphillips:
:+1:

PR #353, opened 2016-03-28
Removed flakey test

MalazAlamir:
Taking out a flakey test that fails because tools.usps.com can't give tracking information for a shipment once its handed to canada post.

Besides the test directly below it more or less tests the same thing.

cc/ @mdking @lucasuyezu @thegedge


thegedge:
:fire: all the flakey tests (or remove their sources of flakiness)!

Agreed that the test below it is basically doing what this test is doing :+1:

cyprusad:
LGTM :shipit:

PR #354, opened 2016-03-29
UPS. Ability to Bill Third Party for Duties and Taxes

eboswort:
This adds the ability to bill duties and taxes to a third party shipping account.

Also contained in this PR is a change to actually use parameter `negotiated_rates` which was already documented above the `create_shipment` method definition. We were inferring that based on the presence of an `origin_account` but given that negotiated rates are based on who's paying, the existence of third party billing means it doesn't really make sense to do that. However this could be a breaking change for anyone relying on the `origin_account` to automatically turn on negotiated rates.
Really I think we should probably rename `origin_account` to be `shipper_account` since that's a better description of what it is.

### DIFF
--- a/lib/active_shipping/carriers/ups.rb
+++ b/lib/active_shipping/carriers/ups.rb
@@ -318,7 +318,7 @@ module ActiveShipping
     # Build XML node to request a shipping label for the given packages.
     #
     # options:
-    # * origin_account: who will pay for the shipping label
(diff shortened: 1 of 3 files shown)


kmcphillips:
Hey @eboswort. Sorry I missed this.

Change looks good! 👍

eboswort:
Awesome! Thanks, I'll merge.

PR #357, opened 2016-04-21
Add ResponseContentError for raised UPS error

jgillman:
Errors from UPS were getting thrown as runtime errors which are harder to catch. This change mirrors how [errors are thrown in the FedEx carrier](https://github.com/Shopify/active_shipping/blob/master/lib/active_shipping/carriers/fedex.rb#L606).


jonathankwok:
Thanks for the fix @jgillman!

PR #359, opened 2016-04-26
Save description for UPS shipment event in message

jonathankwok:
## About
For the other carrier adapters, we always save something for the `message` of a `ShipmentEvent` - 
 - [USPS does it here, storing the event's `Event` in `message`](https://github.com/Shopify/active_shipping/blob/master/lib/active_shipping/carriers/usps.rb#L627)
 - [FedEx also does it here, saving `EventDescription`](https://github.com/Shopify/active_shipping/blob/master/lib/active_shipping/carriers/fedex.rb#L664)
 - [Canada Post saves the `event-description` in `message`](https://github.com/Shopify/active_shipping/blob/master/lib/active_shipping/carriers/canada_post_pws.rb#L328)

UPS doesn't do anything, and just [gives it a `nil` value](https://github.com/Shopify/active_shipping/blob/master/lib/active_shipping/carriers/ups.rb#L853). Rather than `nil` the value, we can do what USPS and FedEx do and pass in the event description.

This change will keep the carriers consistent with each other.

## Changes
Simply store the text in the `'Status/StatusType/Description')` node in `message` for each `ShipmentEvent`.

## Testing
Added another unit test in `test/unit/ups_test.rb` to verify the existence of `message` for each shipment event in a response.

@kmcphillips @mdking @RichardBlair @MalazAlamir

### DIFF
--- a/lib/active_shipping/carriers/ups.rb
+++ b/lib/active_shipping/carriers/ups.rb
@@ -850,7 +850,7 @@ module ActiveShipping
             type_code = activity.at('Status/StatusType/Code').text
             zoneless_time = parse_ups_datetime(:time => activity.at('Time'), :date => activity.at('Date'))
             location = location_from_address_node(activity.at('ActivityLocation/Address'))
-            ShipmentEvent.new(description, zoneless_time, location, nil, type_code)
+            ShipmentEvent.new(description, zoneless_time, location, description, type_code)
           end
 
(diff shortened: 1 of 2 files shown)


RichardBlair:
:shipit: on 💚

mdking:
👍

kmcphillips:
Change looks good! Let's get the build fixed out of band first though.

PR #360, opened 2016-04-26
Canada Post PWS fix for missing tracking-pin (optional field)

evendir:
Per the Canada Post documentation, tracking-pin may not be present in the non-contract-shipping-info response.  Currently that case is resulting in an exception.


kmcphillips:
@jonathankwok 👀 ?

jonathankwok:
Looks good to me, after I merge the fixes for the failing build I'll merge this in.

Thanks for the fix, @evendir!

PR #361, opened 2016-04-26
Use multiple attempted delivery codes for USPS tracking

jonathankwok:
## About
There are multiple possible `attempted_delivery` codes - see page 335 of [this](https://ribbs.usps.gov/intelligentmail_package/documents/tech_guides/PUB199IMPBImpGuide.pdf), or look at this screenshot:

image

There are multiple codes for "Notice Left"/Attempted Delivery - rather than just use code 55, use all of the ones in the appendix. Code H0 came from me finding responses manually through a database of tracking codes, seems to also be Attempted Delivery (unsurprisingly not documented).

## Changes
Add an array of known `ATTEMPTED_DELIVERY_CODES` and check for inclusion when saving the `attempted_delivery_date`.

This fixes part of the red build, which doesn't actually depend on this fix - it was just an outdated tracking number. The other half is fixed by https://github.com/Shopify/active_shipping/pull/362

Review: @kmcphillips @RichardBlair @mdking @MalazAlamir

### DIFF
--- a/lib/active_shipping/carriers/usps.rb
+++ b/lib/active_shipping/carriers/usps.rb
@@ -113,6 +113,8 @@ module ActiveShipping
       :package_service => 'PACKAGESERVICE'
     }
 
+    ATTEMPTED_DELIVERY_CODES = %w(02 53 54 55 56 H0)
+
     # Array of U.S. possessions according to USPS: https://www.usps.com/ship/official-abbreviations.htm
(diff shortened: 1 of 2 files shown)


RichardBlair:
🐑 :it:

MalazAlamir:
🚢

mdking:
Nice find! 👍

PR #362, opened 2016-04-26
Take into account business hours for UPS delivery estimates

jonathankwok:
### Summary
Time zone shiftiness going on:

```
[1] pry(main)> Date.today
=> Tue, 26 Apr 2016
[2] pry(main)> 3.business_days.from_now
=> 2016-05-02 09:00:00 +0000
[3] pry(main)> 3.business_days.after(Date.today)
=> Fri, 29 Apr 2016
```

`from_now` just adds the n days, whereas `.after` takes into account [business hours](https://github.com/bokmann/business_time/blob/develop/lib/business_time/business_days.rb#L28), adding an extra day if the present time is past business hours (which is why it's flaky right now, since most of the time we're running our builds outside of business hours for the system (UTC)).

Carrier responses take into account business time and rolling forward, so we should too in our tests.

Fixes half of the red build - the other half is fixed by https://github.com/Shopify/active_shipping/pull/361
Review: @kmcphillips @RichardBlair @mdking @MalazAlamir

### DIFF
--- a/test/remote/ups_test.rb
+++ b/test/remote/ups_test.rb
@@ -283,7 +283,7 @@ class RemoteUPSTest < Minitest::Test
     assert response.success?
     refute_empty response.delivery_estimates
     ground_delivery_estimate = response.delivery_estimates.select {|de| de.service_name == "UPS Ground"}.first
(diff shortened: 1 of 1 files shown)


RichardBlair:
:shipit:

MalazAlamir:
🚢

PR #365, opened 2016-05-05
Remove build: true from ruby

kmcphillips:
@jonathankwok @thegedge 

```
~/source/active_shipping(master ✔) dev up
🦄  the build: true option for ruby in dev.yml is obsolete and no longer has any effect.
✓ 1/2: Installing Ruby (nothing to do)
✓ 2/2: Running Bundler (nothing to do)
```


jonathankwok:
🚢

PR #374, opened 2016-05-18
Add cancel_shipment to the interface

jonathanmwatson:
We need to add cancel_shipment to the interface. Will create an issue to come back and add in actual functionality to all carriers we have implemented.

I think this is everything I need to get started. Let me know if you have any other context that I'm missing

@jonathankwok @RichardBlair @mdking @garethson @kmcphillips

### DIFF
--- a/lib/active_shipping/carrier.rb
+++ b/lib/active_shipping/carrier.rb
@@ -7,6 +7,7 @@ module ActiveShipping
   #
   # @see #find_rates
   # @see #create_shipment
+  # @see #cancel_shipment
   # @see #find_tracking_info
   #
   # @!attribute test_mode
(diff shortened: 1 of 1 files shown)


lib/active_shipping/carrier.rb
```diff
@@ -67,6 +68,19 @@ def create_shipment(origin, destination, packages, options = {})
...
+    # Cancels a shipment with a carrier.
+    #
+    # @note Override with whatever you need to cancel a shipping label
+    #
+    # @param shipment_id [String] The unique identifier of the shipment to cancel. 
+    # This can be shipment_id or tracking number depending on carrier. Up to you and 
+    # the carrier
+    # @return [ActiveShipping::ShipmentResponse] The response from the carrier. This
+    #   response in most cases has a cancellation id.
+    def cancel_shipment(shipment_id)
```
mdking: Following with the existing pattern of the interface methods can you add a second argument `options = {}`

mdking:
👍

garethson:
Good idea. 👍

jonathankwok:
Looks good. 🚢

PR #373, opened 2016-05-17
specify homebrew 2.2.5 package as the default dev ruby version

rwu1997:
Specified homebrew's 2.2.5 version in dev.yml so `dev` commands work again.

cc: @jonathankwok @MalazAlamir

PR #378, opened 2016-06-02
Bump ruby to 2.3.1

kmcphillips:
@rwu1997 @thegedge @MalazAlamir 

Bumping the ruby version to `2.3.1`.

There's a typo in this package definition and `version` should be plural. From #373. There are bugs in any `2.2.x` versions because of conflicts with the shopify version of ruby. So no harm in bumping it to the most current.

### DIFF
--- a/dev.yml
+++ b/dev.yml
@@ -1,9 +1,7 @@
 name: active_shipping
 
 up:
-  - ruby:
-      package: homebrew/version/ruby22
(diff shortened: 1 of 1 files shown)


kmcphillips:
Also including 2.3 in Travis.

MalazAlamir:
LGTM :shipit:

mikeletscher:
👍

Issue #381, opened 2016-07-02
Rails 5 Stable?


Sk8withaguitar:
Hello, I would like to know if you will be updating the gem to be compatible with Rails 5 stable? Currently in 5 beta and RC the gem is compatible but not for 5.0.0. Thanks

chaserx:
Yes I'd like to know the same. I blindly tried updating the gem to use activesupport 5, but it seems that there's something funky with DateTime objects. I have not investigated further.

```shell
Finished in 8.356243s, 62.7076 runs/s, 177.8311 assertions/s.

  1) Failure:
ShipwireTest#test_successfully_get_domestic_rates [/Users/chase/scratch_paper/active_shipping/test/unit/carriers/shipwire_test.rb:72]:
--- expected
+++ actual
@@ -1 +1 @@
-Mon, 08 Aug 2011 00:00:00 +0000
+Mon, 01 Aug 2011 00:00:00 +0000



  2) Failure:
ShipwireTest#test_successfully_get_international_rates [/Users/chase/scratch_paper/active_shipping/test/unit/carriers/shipwire_test.rb:47]:
--- expected
+++ actual
@@ -1 +1 @@
-Mon, 08 Aug 2011 00:00:00 +0000
+Mon, 01 Aug 2011 00:00:00 +0000



  3) Failure:
UPSTest#test_delivery_range_takes_weekend_into_consideration [/Users/chase/scratch_paper/active_shipping/test/unit/carriers/ups_test.rb:255]:
--- expected
+++ actual
@@ -1 +1 @@
-[nil, 2012-06-15 00:00:05 UTC, 2012-06-15 00:00:04 UTC, 2012-06-15 00:00:03 UTC, 2012-06-15 00:00:03 UTC, 2012-06-15 00:00:03 UTC]
+[nil, Fri, 15 Jun 2012 00:00:00 +0000, Fri, 15 Jun 2012 00:00:00 +0000, Fri, 15 Jun 2012 00:00:00 +0000, Fri, 15 Jun 2012 00:00:00 +0000, Fri, 15 Jun 2012 00:00:00 +0000]
```

kmcphillips:
Hello!

Yeah we are working on this and will release a new version supporting Rails 5 soon.

PR #383, opened 2016-07-04
Add test for base Carrier class

kmcphillips:
@garethson @jonathankwok 

Related to #381

Digging into getting Rails 5 support. We have failing tests around date offsets on the branch. I found the `Carrier` base class had no tests, and I think the bug may be in one of the methods in this class.

Adding tests.

Build is _probably_ going to be red as surrounding the US holiday a bunch of delivery times on the remote tests are wrong. 🇺🇸 🎆

### DIFF
--- a/test/unit/carrier_test.rb
+++ b/test/unit/carrier_test.rb
@@ -1,0 +1,106 @@
+require 'test_helper'
+
+class CarrierTest < Minitest::Test
+  class ExampleCarrier < Carrier
+    cattr_reader :name
+    @@name = "Example Carrier"
+  end
+
+  def setup
+    @carrier = ExampleCarrier.new
+  end
+
+  def test_find_rates_not_implemented
+    assert_raises NotImplementedError do
+      @carrier.find_rates(nil, nil, nil)
+    end
+  end
+
+  def test_create_shipment_not_implemented
+    assert_raises NotImplementedError do
+      @carrier.create_shipment(nil, nil, nil)
+    end
+  end
+
+  def test_cancel_shipment_not_implemented
+    assert_raises NotImplementedError do
+      @carrier.cancel_shipment(nil)
+    end
+  end
+
+  def test_find_tracking_info_not_implemented
+    assert_raises NotImplementedError do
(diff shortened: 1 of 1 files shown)


kmcphillips:
FWIW:

```
~/source/active_shipping(test-carrier-base ✔) dev test test/unit/carrier_test.rb
Run options: --seed 38102

# Running:

.............

Finished in 0.005136s, 2531.1753 runs/s, 4283.5275 assertions/s.

13 runs, 22 assertions, 0 failures, 0 errors, 0 skips
```

garethson:
Thank you. 👏 

Were the failing test Rails 5 only ... failures? Or existing soft failures?

test/unit/carrier_test.rb
```diff
@@ -0,0 +1,100 @@
...
+
+  def test_save_request
+    request = Object.new
+    assert_nil @carrier.last_request
+    @carrier.send(:save_request, request)
+    assert_equal request, @carrier.last_request
+  end
+
+  def test_timestamp_from_business_day_returns_two_days_in_the_future
+    current = DateTime.civil(2016, 7, 19) # Tuesday
```
jonathankwok: TIL of `civil`.
Interestingly, [`civil`](http://ruby-doc.org/stdlib-1.9.3/libdoc/date/rdoc/DateTime.html#method-c-civil) is an alias to [`.new`](http://ruby-doc.org/stdlib-1.9.3/libdoc/date/rdoc/DateTime.html#method-c-new).

kmcphillips:
Yeah, we get three date related failures with `activesupport 5.0.0` in the gemfile. Still digging into that, but want test coverage to help me do that.

test/unit/carrier_test.rb
kmcphillips: Interesting. This is a copy/pasta from another test. But I can defs change it to `new` as that's clearer.

jonathankwok:
👍  to me

kmcphillips:
Merging despite the 🔴 since the fails are remote tests, and this is all about getting our build passing correctly again.

PR #384, opened 2016-07-04
Shipwire delivery date test cleanup

kmcphillips:
@garethson @jonathankwok 

Related to #381 and follows #383. 

Our Rails 5 test failures are all involved in date arithmetic and Shipwire. This does a few things to clean up how we test dates and delivery estimates to continue to narrow in on the problem.

### DIFF
--- a/test/unit/carriers/shipwire_test.rb
+++ b/test/unit/carriers/shipwire_test.rb
@@ -24,67 +24,65 @@ class ShipwireTest < Minitest::Test
 
   def test_successfully_get_international_rates
     date = DateTime.new(2011, 8, 1)
-    DateTime.expects(:now).returns(date).at_least_once
     @carrier.expects(:ssl_post).returns(xml_fixture('shipwire/international_rates_response'))
 
-    response = @carrier.find_rates(
-                 location_fixtures[:ottawa],
-                 location_fixtures[:london],
-                 package_fixtures.values_at(:book, :wii),
-                 :order_id => '#1000',
-                 :items => @items
-               )
+    Timecop.freeze(date) do
+      response = @carrier.find_rates(
+                   location_fixtures[:ottawa],
+                   location_fixtures[:london],
+                   package_fixtures.values_at(:book, :wii),
+                   :order_id => '#1000',
+                   :items => @items
+                 )
 
-    assert response.success?
+      assert response.success?
 
-    assert_equal 1, response.rates.size
+      assert_equal 1, response.rates.size
 
-    assert international = response.rates.first
-    assert_equal "INTL", international.service_code
-    assert_equal "UPS", international.carrier
-    assert_equal "UPS Standard", international.service_name
-    assert_equal 2806, international.total_price
-    assert_equal date + 7, international.delivery_date
-    assert_equal [date + 1, date + 7], international.delivery_range
+      assert international = response.rates.first
+      assert_equal "INTL", international.service_code
+      assert_equal "UPS", international.carrier
+      assert_equal "UPS Standard", international.service_name
+      assert_equal 2806, international.total_price
+      assert_equal [date + 1.day, date + 7.days], international.delivery_range
+    end
   end
 
   def test_successfully_get_domestic_rates
     date = DateTime.new(2011, 8, 1)
-    DateTime.expects(:now).returns(date).at_least_once
     @carrier.expects(:ssl_post).returns(xml_fixture('shipwire/rates_response'))
 
-    response = @carrier.find_rates(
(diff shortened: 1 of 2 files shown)


test/unit/carriers/shipwire_test.rb
```diff
@@ -24,67 +24,65 @@ def test_response_with_no_rates_is_unsuccessful
...
-    assert_equal 1, response.rates.size
-
-    assert international = response.rates.first
-    assert_equal "INTL", international.service_code
-    assert_equal "UPS", international.carrier
-    assert_equal "UPS Standard", international.service_name
-    assert_equal 2806, international.total_price
-    assert_equal date + 7, international.delivery_date
-    assert_equal [date + 1, date + 7], international.delivery_range
+    Timecop.freeze(date) do
```
kmcphillips: Use `Timecop` rather than stubbing.

test/unit/carriers/shipwire_test.rb
```diff
@@ -24,67 +24,65 @@ def test_response_with_no_rates_is_unsuccessful
...
+      assert response.success?
+
+      assert_equal 1, response.rates.size
+
+      assert international = response.rates.first
+      assert_equal "INTL", international.service_code
+      assert_equal "UPS", international.carrier
+      assert_equal "UPS Standard", international.service_name
+      assert_equal 2806, international.total_price
+      assert_equal [date + 1.day, date + 7.days], international.delivery_range
```
kmcphillips: Call `.days` rather than adding a scalar, which only works in _some_ cases.

test/unit/rate_estimate_test.rb
```diff
@@ -58,4 +58,8 @@ def test_delivery_category_is_set
     assert_equal "local_delivery", est.delivery_category
   end
 
+  def test_delivery_date_pulls_from_delivery_range
+    assert_equal [DateTime.new(2016, 7, 1), DateTime.new(2016, 7, 3)], @rate_estimate.delivery_range
+    assert_equal DateTime.new(2016, 7, 3), @rate_estimate.delivery_date
```
kmcphillips: Test `delivery_date` as being `delivery_range.last` in all cases, meaning we don't have to test `delivery_date` in other tests. Makes the tests simpler and removes duplication.

test/unit/carriers/shipwire_test.rb
```diff
@@ -24,67 +24,65 @@ def test_response_with_no_rates_is_unsuccessful
...
-    assert response.success?
-
-    assert_equal 1, response.rates.size
-
-    assert international = response.rates.first
-    assert_equal "INTL", international.service_code
-    assert_equal "UPS", international.carrier
-    assert_equal "UPS Standard", international.service_name
-    assert_equal 2806, international.total_price
-    assert_equal date + 7, international.delivery_date
```
kmcphillips: Don't test `delivery_date` as we now have a test asserting it will always be `delivery_range.last`

test/unit/carriers/shipwire_test.rb
jonathankwok: Interesting that this doesn't work in all cases, since `date + n` where `n` is an integer and `date` is a `DateTime` should result in adding one day.

Either way, explicitness is good here.

jonathankwok:
👍

test/unit/carriers/shipwire_test.rb
kmcphillips: Behaviour depends on the class.

```
irb(main):011:0> time = Time.now
=> 2016-07-04 14:38:02 -0400
irb(main):012:0> datetime = time.to_datetime
=> #<DateTime: 2016-07-04T14:38:02-04:00 ((2457574j,67082s,766271000n),-14400s,2299161j)>
irb(main):013:0> time + 1
=> 2016-07-04 14:38:03 -0400
irb(main):014:0> datetime + 1
=> #<DateTime: 2016-07-05T14:38:02-04:00 ((2457575j,67082s,766271000n),-14400s,2299161j)>
```

One of our Rails 5 bugs is around returning `DateTime` vs `Time` when doing time math.

kmcphillips: We should always be explicit anyway.

test/unit/rate_estimate_test.rb
```diff
@@ -2,11 +2,11 @@
 
 class RateEstimateTest < Minitest::Test
```
garethson: #yolo

garethson:
Awesome. Thanks.

kmcphillips:
Same 🔴 with remotes. Iterating.

PR #385, opened 2016-07-04
Correctly offset days for business days

kmcphillips:
@garethson @jonathankwok 

Works toward #381. Our test failures in Rails 5 are around date math, and the `Time` vs `DateTime` calculations.

This fixes assumptions about class, and cleans up some really awkward ruby syntax. Tests the weekend cases too.

### DIFF
--- a/lib/active_shipping/carrier.rb
+++ b/lib/active_shipping/carrier.rb
@@ -72,8 +72,8 @@ module ActiveShipping
     #
     # @note Override with whatever you need to cancel a shipping label
     #
-    # @param shipment_id [String] The unique identifier of the shipment to cancel. 
-    #  This can be shipment_id or tracking number depending on carrier. Up to you and 
+    # @param shipment_id [String] The unique identifier of the shipment to cancel.
+    #  This can be shipment_id or tracking number depending on carrier. Up to you and
     #  the carrier
     # @param options [Hash] Carrier-specific parameters.
     # @return [ActiveShipping::ShipmentResponse] The response from the carrier. This
@@ -162,12 +162,15 @@ module ActiveShipping
     def timestamp_from_business_day(days)
       return unless days
       date = DateTime.now.utc
+
       days.times do
-        begin
-          date = date + 1
-        end while [0, 6].include?(date.wday)
+        date += 1.day
+
+        date += 2.days if date.saturday?
+        date += 1.day if date.sunday?
       end
-      date
+
+      date.to_datetime
     end
   end
 end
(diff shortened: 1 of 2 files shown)


lib/active_shipping/carrier.rb
```diff
@@ -162,12 +162,15 @@ def save_request(r)
     def timestamp_from_business_day(days)
       return unless days
       date = DateTime.now.utc
+
       days.times do
-        begin
-          date = date + 1
-        end while [0, 6].include?(date.wday)
```
garethson: :/

test/unit/carrier_test.rb
```diff
@@ -84,6 +84,24 @@ def test_timestamp_from_business_day_returns_fifteen_days_in_the_future
...
+  def test_timestamp_from_business_day_handles_saturday
+    current = DateTime.new(2016, 7, 9) # Saturday
+    expected = DateTime.new(2016, 7, 11)
+
+    Timecop.freeze(current) do
+      assert_equal expected, @carrier.send(:timestamp_from_business_day, 1)
+    end
+  end
+
+  def test_timestamp_from_business_day_handles_saturday
```
jonathankwok: test name duplicate from above.

test/unit/carrier_test.rb
```diff
@@ -84,6 +84,24 @@ def test_timestamp_from_business_day_returns_fifteen_days_in_the_future
     end
   end
 
+  def test_timestamp_from_business_day_handles_saturday
+    current = DateTime.new(2016, 7, 9) # Saturday
+    expected = DateTime.new(2016, 7, 11)
+
+    Timecop.freeze(current) do
+      assert_equal expected, @carrier.send(:timestamp_from_business_day, 1)
```
garethson: I guess this is a private method?

test/unit/carrier_test.rb
kmcphillips: Nice catch. I wonder what happens.

We should convert this whole damn project to the `test "" do` syntax.

test/unit/carrier_test.rb
kmcphillips: It's a helper function from within the `Carrier` class. Used by subclasses. Actually `protected`.

Alternately I could use a delegator from within the example class here, but I think that's less clear.

test/unit/carrier_test.rb
jonathankwok: It should fail and complain about duplicate test existing already.

Also +1 to converting everything to `test "" do`. Doing it on a file-by-file basis would probably be more manageable.

jonathankwok:
:shipit:

test/unit/carrier_test.rb
kmcphillips: It's actually an active support feature, but this gem depends on it explicitly so that should be fine.

PR #387, opened 2016-07-09
ShipmentOption class for Carriers

richardwu:
#### Motive

Most carriers allow for shipment options to be queried for in a rate request or added on a label purchase. We should define an interface for `ShipmentOption` that will make it dead simple to retrieve `name` and `description` as well as filter `option_args` just from the `code` for a given `ShipmentOption`.
#### How to use

Simply subclass `ShipmentOption` with your specific implementation for a given carrier (e.g. `UspsShipmentOption`). You will need to declare `@@carrier_name` and `@@all_options_details` in your subclass.

`@@carrier_name` is self explanatory (name of the carrier). `@@all_options_details` contains a mapping of all the carrier's supported shipment options with details such as `name`, `description` and parameters it requires for the query (`option_params`).

e.g. 

```
 @@all_options_details = {
   'signature_required' => {
     name: 'Signature Required',
     description: 'A signature is required on delivery.',
     option_params: [:value, :fragile]
   }
 }

```

@mdking @mondayrain 
cc: @RichardBlair


kmcphillips:
👍

mdking:
As discussed offline I think some more context around:

- How this class is to be used with respect to `find_rates` and `create_shipment`
- This class should promote a set of common shipment options codes to generalize options across all carriers. E.g. USPS, UPS, Canada Post, etc... all have a form of signature required we should establish a standard code for that IMO and individual adapters can map the carrier specific values to it.

maartenvg:
Closing old PRs that are not going to ship.

PR #388, opened 2016-07-12
Add accessor for CanadaPostPWS @@name

dangerdogz:
Add a cattr_reader to access the @@name variable provided by the class.
This is something standard in the other carriers that is missing from Canada Post Pws.


kmcphillips:
Thanks!

PR #390, opened 2016-07-18
Added shipment_options attribute to RateEstimate

rwu1997:
#### Motivation

There is currently no attribute in `RateEstimate` that allows for OPTIONAL priced options.

#### Solution

Allow for an attribute `shipment_options` to be initialized in `RateEstimate` that can be an `Array` of objects `{ code: String, price: Integer }`. 

Review: @kmcphillips @lucasuyezu

### DIFF
--- a/lib/active_shipping/rate_estimate.rb
+++ b/lib/active_shipping/rate_estimate.rb
@@ -101,7 +101,7 @@ module ActiveShipping
       self.delivery_date = @delivery_range.last
       self.insurance_price = options[:insurance_price]
       self.delivery_category = options[:delivery_category]
-      self.shipment_options = options[:shipment_options]
+      self.shipment_options = options[:shipment_options] || []
     end
 
     # The total price of the shipments in cents.
--- a/lib/active_shipping/rate_estimate.rb
+++ b/lib/active_shipping/rate_estimate.rb
@@ -66,13 +66,19 @@ module ActiveShipping
   # @!attribute delivery_category
   #   The general classification of the delivery method
   #   @return [String]
+  #
+  # @!attribute shipment_options
+  #   Additional priced options bundled with the given rate estimate with price in cents
+  #   @return [Array<{ code: String, price: Integer }>]
+  #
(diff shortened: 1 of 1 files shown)


lib/active_shipping/rate_estimate.rb
```diff
@@ -95,6 +101,7 @@ def initialize(origin, destination, carrier, service_name, options = {})
       self.delivery_date = @delivery_range.last
       self.insurance_price = options[:insurance_price]
       self.delivery_category = options[:delivery_category]
+      self.shipment_options = options[:shipment_options]
```
kmcphillips: Want to default this to be an empty array? 

It's the difference between having to `if rate_estimate.shipment_options` guard iteration and things.

kmcphillips: `|| []` I mean.

lib/active_shipping/rate_estimate.rb
```diff
@@ -66,13 +66,19 @@ module ActiveShipping
   # @!attribute delivery_category
   #   The general classification of the delivery method
   #   @return [String]
+  #
+  # @!attribute shipment_options
+  #   Additional priced options bundled with the given rate estimate with price in cents
+  #   @return [Array<{ code: String, price: Integer }>]
```
kmcphillips: Certain this is all you'll need? No convenience methods or checks that would merit an `ActiveShipping::ShippingOption` class? Using an array of hashes in this case is in line with what exists in this gem. But just making sure we are forward thinking.

kmcphillips:
I'll release this today.

lib/active_shipping/rate_estimate.rb
rwu1997: The most optimal case is to use a `ShipmentOption` class https://github.com/Shopify/active_shipping/pull/387 but the interface is still ambiguous at this point.

An inline object will allow us to iterate on the interface quickly until we're confident with what attributes belongs in `ShipmentOption`.

Issue #392, opened 2016-07-27
Support Various FedEx Label Formats


irongaze:
UPS and other carriers allow overriding the default label format returned when creating shipments.  FedEx is currently hard-coded to PNG, which is not a particularly printable format.  My consultancy needs a fix for this, and I'm opening this issue to request feedback from the core team about how best to implement it.

The simplest solution would be to add an option to the carrier's possible option set (say, :label_format to be consistent with UPS).  This would be a one-liner.  

There are additional possible levels of error checking (validating the format is a valid FedEx format) as well as the possibility of having a cross-carrier approach (for example, specifying 'zpl' translates to 'ZPLII' on FedEx, 'ZPL' on UPS, etc).

What's the team's thoughts on this?  Are we moving towards a collection of carriers, or a shipping abstraction layer?

Regardless, I'll fork and get the one-liner ready.

irongaze:
Pull request created: #393 

Went with the minimum fix.

irongaze:
PR merged, closing issue.  Use :label_format in label creation options for a good time...

PR #393, opened 2016-08-01
Support various FedEx label formats - issue #392

irongaze:
Simple fix with tests to enable specifying what format a generated label should be returned in via the FedEx API.

As a side-note, what the hell is going on with the changelog file - where are the changes for version 1.6.2 until now?  What's up with all the remote test failures in the test suite?

### DIFF
--- a/lib/active_shipping/carriers/fedex.rb
+++ b/lib/active_shipping/carriers/fedex.rb
@@ -136,6 +136,9 @@ module ActiveShipping
 
     DEFAULT_LABEL_STOCK_TYPE = 'PAPER_7X4.75'
 
+    # Available return formats for image data when creating labels
+    LABEL_FORMATS = ['DPL', 'EPL2', 'PDF', 'ZPLII', 'PNG'] 
+
     def self.service_name_for_code(service_code)
       SERVICE_TYPES[service_code] || "FedEx #{service_code.titleize.sub(/Fedex /, '')}"
     end
@@ -219,7 +222,7 @@ module ActiveShipping
 
             xml.LabelSpecification do
               xml.LabelFormatType('COMMON2D')
-              xml.ImageType('PNG')
+              xml.ImageType(options[:label_format] || 'PNG')
               xml.LabelStockType(options[:label_stock_type] || DEFAULT_LABEL_STOCK_TYPE)
             end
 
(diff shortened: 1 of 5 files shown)


kmcphillips:
Related to #392

kmcphillips:
Changelog we have failed to keep up to date apparently! I'll go back and backfill it.

Remote test failures are a _constant_ game of cat and mouse. Often they fail when there is a holiday or things like that. We're always working to improve them.

test/remote/kunaki_test.rb
```diff
@@ -19,8 +19,8 @@ def test_successful_rates_request
                )
 
     assert response.success?
-    assert_equal 3, response.rates.size
-    assert_equal ["UPS 2nd Day Air", "UPS Ground", "UPS Next Day Air Saver"], response.rates.collect(&:service_name).sort
+    assert_equal 4, response.rates.size
+    assert_equal ["UPS 2nd Day Air", "UPS Ground", "UPS Next Day Air Saver", "USPS First Class Mail"], response.rates.collect(&:service_name).sort
```
kmcphillips: This was to fix a remote failure?

lib/active_shipping/carriers/fedex.rb
```diff
@@ -136,6 +136,9 @@ class FedEx < Carrier
 
     DEFAULT_LABEL_STOCK_TYPE = 'PAPER_7X4.75'
 
+    # Available return formats for image data when creating labels
+    LABEL_FORMATS = ['DPL', 'EPL2', 'PDF', 'ZPLII', 'PNG']
```
kmcphillips: Are these consumed? Why does this exist? To be able to check `ActiveShipping::Fedex::LABEL_FORMATS` outside of the gem?

If so, it should have a quick test.

irongaze: More as a quick reference for API consumers so they don't have to wade through the specs.  I considered a more robust API here (input validation, eg), but I'm following the general coding practices of the gem which are to be very loose on the inputs accepted.

test/remote/kunaki_test.rb
irongaze: Yes.  Whole Kunaki remote suite was failing due to more options being returned.

lib/active_shipping/carriers/fedex.rb
kmcphillips: Ok, I'm good with this.

kmcphillips:
Released `1.8.2`

PR #394, opened 2016-08-03
Add description to rate estimates

christianblais:
This adds a description field to rate estimates so that carriers
can better describe provided rates. API Carriers already support this within Shopify,
but I believe the API should be the same across the board.

### DIFF
--- a/lib/active_shipping/rate_estimate.rb
+++ b/lib/active_shipping/rate_estimate.rb
@@ -27,6 +27,10 @@ module ActiveShipping
   #   The code of the shipping service
   #   @return [String]
   #
+  # @!attribute description
+  #   Public description of the shipping service (e.g. "2 days delivery")
+  #   @return [String]
+  #
   # @!attribute shipping_date
   #   The date on which the shipment will be expected. Normally, this means that the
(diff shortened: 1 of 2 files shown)


christianblais:
@kmcphillips @richardwu for review

richardwu:
👍

christianblais:
Would you guys know why it fails CI? Doesn't look relates to my change at all.

richardwu:
Flaky remote tests. Just have to rebuild and 🙏 until they're green.

christianblais:
Heuh, none of that test suite looks legit. There's tons of warnings about undefined variables, which basically avoid running all the test. As an example, [here](https://github.com/Shopify/active_shipping/blob/7c161f698edc0975730ff1783891b1e7f4d8d713/test/unit/rate_estimate_test.rb#L11), `@service_name` appears to be undefined, hence cancelling the entire test suite.

christianblais:
Finally got a green :-)

jonathankwok:
Looks good to me. 👍

kmcphillips:
Released in `1.8.3`

PR #395, opened 2016-08-17
RequestError for requests that do not adhere to the API

richardwu:
### Context

It is expected that a client of an `active_shipping` API passes valid AND "business-logic correct" arguments.

E.g. An destination being passed in through `find_rates` should be a valid address, or shipment options codes being passed in must be supported by the API.

If not, a `RequestError` can be raised to highlight the incorrect request.

Review: @kmcphillips @thegedge 
cc: @RichardBlair @mdking

### DIFF
--- a/lib/active_shipping/errors.rb
+++ b/lib/active_shipping/errors.rb
@@ -2,6 +2,9 @@ module ActiveShipping
   class Error < ActiveUtils::ActiveUtilsError
   end
 
+  class RequestError < ActiveShipping::Error
+  end
+
   class ResponseError < ActiveShipping::Error
     attr_reader :response
 


kmcphillips:
Do you need extra metadata like the `ResponseError`? Why do you need an exception different from the base exception?

richardwu:
The main motivation of making the distinction of `RequestError` is so that the client knows there's something wrong with the request parameters that's not a generic Ruby `ArugmentError`.

We could probably pass in the offending `parameters` and corresponding `messages`. We don't enough use cases right now to know what's right, but we could also introduce additional metadata.

Here's some context from Shipify https://github.com/Shopify/shipify/pull/1231#discussion_r75152727.

RichardBlair:
@richardwu because this is an open source project could you try to explain, in a generic way why this is needed?

thegedge:
I think it might be worthwhile to adjust existing exceptions that potentially fit under this new category of exception. For example, do any of [these](https://github.com/Shopify/active_shipping/blob/b501556077e3ea435cc88acc14e4ab6dfba9ab19/lib/active_shipping/carriers/canada_post_pws.rb#L889-L892) make sense as a `RequestError`?

lib/active_shipping/errors.rb
```diff
@@ -2,6 +2,9 @@ module ActiveShipping
   class Error < ActiveUtils::ActiveUtilsError
   end
 
+  class RequestError < ActiveShipping::Error
```
thegedge: `ActiveShipping::Error` is an `ActiveUtils::ActiveUtilsError`, which is an alias for `HTTPRequestError`. Not really appropriate, but I thought I'd point it out.

richardwu: We could probably just change `ActiveShipping::Error` to inherit directly from `StandardError`, no?

What's the benefit of `ActiveUtils::ActiveUtilsError`?

thegedge: Not sure if there's any real benefit, but maybe someone else knows of one.

mdking:
Exceptions like error codes increase the footprint of the interface. I think one of the reason this is being proposed is to have a base exception to indicate an error occurred prior to sending the actual request apart from using `ResponseError`. 

@thegedge based on the exceptions you linked to 3 of the 4 could inherit from this new exception class and leverage any new functionality added beyond what `StandardException` provides.

richardwu:
Closing for now, re-opening in another PR to refactor all the errors.

PR #396, opened 2016-08-22
ActiveShipping::Error inherit from StandardError vs. ActiveUtilsError, introduce RequestError

richardwu:
### Motivation

Following the discussion in https://github.com/Shopify/active_shipping/pull/395, it was deemed convenient to introduce an `ActiveShipping::RequestError` to be raised when an input is invalid before a request is made to an external 3rd party.

It was also brought up that `ActiveShipping::Error` inherits from `ActiveUtils::ActiveUtilsError` which is an alias for `HttpRequestError` (https://github.com/Shopify/active_shipping/pull/395#discussion_r75489692). Since some of active_shipping raises `ActiveShipping::Error` irregardless of the actual `HttpRequest`, it would make more sense if `ActiveShipping::Error` inherited from `StandardError` instead.

This change may break `rescue` implementations of `ActiveShipping::Error`, so we should probably do a major version bump.

Review: @kmcphillips @mdking @thegedge


richardwu:
Probably won't be able to push this out before I leave.

Feel free to take this 👍

maartenvg:
Closing old PRs that are not going to ship.

Issue #371, opened 2016-05-17
USPS Address Validation API


Tonkpils:
Going off https://github.com/Shopify/active_shipping/issues/39, would active_shipping be open to adding a method for the address validation API for USPS carrier? 

I'd be happy to make a PR for this but wanted to get your +1 before moving ahead.

jonathankwok:
Hey! I'd be interested in seeing an optional method that allows us to validate addresses. The key is to make it not a required component of a typical request, since adding an extra API request would slow down the process.

Feel free to create a PR for this, I'd be happy to review it.

alecharmon:
@Tonkpils / @jonathankwok   (I dug around in closed pr's and couldn't find this issue referenced ) I would be happy to pick this up if there is still a need

alecharmon:
Hey so rather then validate where there is a carrier method that returns a boolean, I think it makes more sense to return what USPS interpreted the correct location to be and nil if it can't be validated.

PR #406, opened 2016-11-07
(wip) USPS Address Validation API #371

alecharmon:
## Summary 
First stab a pr for #371 , the USPS api is a little odd, there are default address, non-verified, address, and verified addresses. my approach was to base everything off of default address. if it exists that is equivalent imo to verified.
## Example
```carrier.is_valid_address(some_usa_add)```

### CODE REVIEW

File: lib/active_shipping/carriers/usps.rb [hunk 1]
```diff
@@ -304,6 +319,38 @@ def build_tracking_batch_request(tracking_infos, options)
...
+      EOF
+    end
+
+    #return deafault address from USPS
+    def parse_address_validation_response(response)
+      xml = Nokogiri.XML(response)
+      return nil if xml.at_xpath("//Error")
+      address = xml.at_xpath('//Address')
+      return Location.new(
+        :country => 'US',
```
kmcphillips: Use the new hash syntax please. 1.8.7 is no longer supported.

alecharmon:
(back from the grave 🚶 💀 ) hey, so I did some digging and found that issue with the failing test on Travis CI was caused by calling the USPS Verify. The login name for the api setup on the CI server is not setup for use with the verify API. Here is the raw XML response:
```
"<?xml version=\"1.0\" encoding=\"UTF-8\"?>\n<Error><Number>80040B1A</Number><Description>API Authorization failure. User 677JADED7283 is not authorized to use API Verify.</Description><Source>USPSCOM::DoAuth</Source></Error>"
```
@kmcphillips do you think we could change the username?

PR #407, opened 2016-11-08
remove obsolete todo item

shime:
seems like you figured out how USPS likes to say Ivory Coast


kmcphillips:
How is that?

shime:
It's in the `COUNTRY_NAME_CONVERSIONS` hash: https://github.com/Shopify/active_shipping/blob/7ea03627b9b609a3a59172a99097cd77cf177f4b/lib/active_shipping/carriers/usps.rb#L127

PR #408, opened 2016-11-08
UPS ImportControl support

megetron:
To import a package to your own country you have to use ImportControl with UPS.

When using the ImportControl option, you will also have to add a package description which used for returns so I used it to support ImportControl also.

### DIFF
--- a/lib/active_shipping/carriers/ups.rb
+++ b/lib/active_shipping/carriers/ups.rb
@@ -1076,6 +1076,7 @@ module ActiveShipping
       return nil unless address
       country = address.at('CountryCode').try(:text)
       country = 'US' if country == 'ZZ' # Sometimes returned by SUREPOST in the US
+      country = 'XK' if country == 'KV' # ActiveUtils now refers to Kosovo by XK
       Location.new(
         :country     => country,
         :postal_code => address.at('PostalCode').try(:text),
--- a/lib/active_shipping/carriers/ups.rb
+++ b/lib/active_shipping/carriers/ups.rb
@@ -127,6 +127,10 @@ module ActiveShipping
 
     IMPERIAL_COUNTRIES = %w(US LR MM)
 
+    COUNTRY_MAPPING = {
+      'XK' => 'KV'
+    }.freeze
+
     DEFAULT_SERVICE_NAME_TO_CODE = Hash[UPS::DEFAULT_SERVICES.to_a.map(&:reverse)]
     DEFAULT_SERVICE_NAME_TO_CODE['UPS 2nd Day Air'] = "02"
     DEFAULT_SERVICE_NAME_TO_CODE['UPS 3 Day Select'] = "12"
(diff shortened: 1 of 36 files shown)


lib/active_shipping/carriers/ups.rb
```diff
@@ -660,11 +676,8 @@ def build_package_node(xml, package, options = {})
       xml.Package do
         # not implemented:  * Shipment/Package/PackagingType element
 
-        #return requires description
-        if options[:return]
```
kmcphillips: Why was this check for `return` removed?

lib/active_shipping/carriers/ups.rb
```diff
@@ -376,10 +384,18 @@ def build_shipment_request(origin, destination, packages, options={})
             # Required element. The company whose account is responsible for the label(s).
             build_location_node(xml, 'Shipper', shipper, options)
 
-            if options[:saturday_delivery]
-              xml.ShipmentServiceOptions do
+            xml.ShipmentServiceOptions do
```
kmcphillips: If `saturday_delivery` and `label_method_code` are blank then we shouldn't include this node.

lib/active_shipping/carriers/ups.rb
```diff
@@ -376,10 +384,18 @@ def build_shipment_request(origin, destination, packages, options={})
...
+              if options[:saturday_delivery]
                 xml.SaturdayDelivery
               end
+
+              if (code = options[:label_method_code])
+                xml.ImportControlIndicator
+                xml.LabelMethod do
+                  xml.Code(code)
+                  xml.Description(LABEL_METHOD_CODES[code])
```
kmcphillips: This doesn't guard against an unknown code.

megetron:
http://idaho-dental-studio.com/wp-content/uploads/2012/11/Shipping-Package-XML-Developers-Guide.pdf
Check this and let me know if you have any further comments.

kmcphillips:
I asked a number of questions already. And I'm not searching a 260 page document to get the answer to my question. Please provide the supporting details to explain your changes.

megetron:
Ooops, I merged also another change that I work for fedex. can you reset the fedex.rb file from the pull request? or please tell how can I do it my self?


Anyway, I made the changes.


And regarding the info,
in order to import international package with UPS u will need to add support in request...
According to UPS (page 126 in the file I sent)

**ImportControlIndicator**: Indicates that the Shipment is an ImportControl shipment.
**LabelMethod**:  Type of ImportControl Label
**Code:** LabelMethod types:'01' =
ImportControl Print and
Mail'02' = ImportControl
One-Attempt'03' =
ImportControl ThreeAttempt'04'
= ImportControl
Electronic Label'05' =
ImportControl Print Label
**Description**: Label Method description


Regarding the removal of the [:return] condition, ImportLabel requires package description too.
I think that if you have a package description it can do no harm, so remove this condition. please correct me if I got it wrong.

kmcphillips:
> can you reset the fedex.rb file from the pull request? or please tell how can I do it my self?

I'm sorry, I don't know what this means.

kmcphillips:
Also, the file you linked is four years old, and is not coming from UPS.

maartenvg:
Closing due to inactivity & state of PR.

Issue #389, opened 2016-07-13
NotImplementedError: #create_shipment is not supported by USPS.


jburks-bba:
Trying to create shipments and labels for USPS and I get a NotImplementedError. Any plans to implement create_shipment for USPS?

kmcphillips:
Nope, nothing planned for a direct to USPS integration.

shime:
@kmcphillips would you accept a PR that added that?

kmcphillips:
@shime Sure thing. If it's well tested, supported documentation, etc..

Issue #379, opened 2016-06-03
No usps create shipment method?


shopshow:
For usps carrier, how to create shipment and get printable label (png or pdf)?

PR #411, opened 2016-11-10
add support for USPS label retrieval

shime:
This adds support for retrieving USPS domestic and international labels.

Related issues: #389, #379


zeiv:
:shipit:

test/remote/usps_test.rb
```diff
@@ -125,6 +125,70 @@ def test_international_rates
...
+        location_fixtures[:beverly_hills_with_name],
+        package_fixtures.values_at(:book),
+        :test => true,
+        :service => :first_class
+      )
+    end
+  end
+
+  def test_domestic_label_to_name_errors
+    assert_raises(ResponseError, "To Name is required and may not exceed 100 characters.") do
```
jonathankwok: `assert_raises` doesn't actually verify that the string you pass in is the exception message - it's actually the assertion failure message. You can see it being `pop`'d from `*exp` here: `https://github.com/seattlerb/minitest/blob/master/lib/minitest/assertions.rb#L319`

It's rather misleading, so to verify an exception message what you want to do is actually set the return value of `assert_raises` to a variable, which will hold the exception message. From there, you can call something like `assert_equal "your exception message", exception.message`.

test/remote/usps_test.rb
```diff
@@ -125,6 +125,70 @@ def test_international_rates
...
+      @carrier.create_shipment(
+        location_fixtures[:new_york_with_name],
+        location_fixtures[:beverly_hills],
+        package_fixtures.values_at(:book),
+        :test => true,
```
jonathankwok: You can use the 1.9 colon hash symbol key style here, and throughout the test.

lib/active_shipping/carriers/usps.rb
```diff
@@ -336,6 +408,171 @@ def canned_address_verification_works?
...
+
+      save_request(xml_builder.to_xml)
+    end
+
+    def build_us_location_node(xml, prefix, address)
+      array = [
+        ['Name', address.name],
+        ['Firm', address.company],
+        ['Address1', address.address1],
+        ['Address2', address.address2 || address.address1],
```
jonathankwok: Any reason why we need to duplicate `address1` in the `Address2` field rather than leave it empty?

lib/active_shipping/carriers/usps.rb
```diff
@@ -113,10 +140,22 @@ class USPS < Carrier
...
+      :dangerous_goods => 'DANGEROUSGOODS',
+      :cremated_remains => 'CrematedRemains',
+      :nonnegotiable_document => 'NonnegotiableDocument'
+    )
+
     ATTEMPTED_DELIVERY_CODES = %w(02 53 54 55 56 H0)
 
     # Array of U.S. possessions according to USPS: https://www.usps.com/ship/official-abbreviations.htm
     US_POSSESSIONS = %w(AS FM GU MH MP PW PR VI)
+    DOMESTIC_CODES = US_POSSESSIONS + ['US', nil]
```
jonathankwok: Why would we consider `nil` to be a domestic code?

lib/active_shipping/carriers/usps.rb
```diff
@@ -336,6 +408,171 @@ def canned_address_verification_works?
...
+
+          xml.Size(size_code)
+        end
+      end
+
+      save_request(xml_builder.to_xml)
+    end
+
+    def build_us_location_node(xml, prefix, address)
+      array = [
```
jonathankwok: Bit of a nitpick but for clarity's sake, naming a variable something a little more descriptive than the data type is almost always helpful. `address_array` or `location_array` would make subsequent operations on this array much easier to understand at a glance.

lib/active_shipping/carriers/usps.rb
```diff
@@ -336,6 +408,171 @@ def canned_address_verification_works?
...
+
+          build_world_location_node(xml, 'From', origin)
+          build_world_location_node(xml, 'To', destination)
+
+          size_code = USPS.size_code_for(package)
+          container = CONTAINERS[package.options[:container]]
+          container ||= (package.cylinder? ? 'NONRECTANGULAR' : 'RECTANGULAR') if size_code == 'LARGE'
+          xml.Container(container)
+
+          xml.ShippingContents do |xml|
```
jonathankwok: This doesn't seem right - this is supposed to be representative of the customs declaration.

zeiv:
Regarding manual testing, yes, we are currently using this branch in production and can get labels without a hitch

lib/active_shipping/carriers/usps.rb
zeiv: The USPS API is _really_ picky.  If the street address is a single line, it expects it to be in the Address2 field.

lib/active_shipping/carriers/usps.rb
shime: To assume US when a package doesn't have the country set.

test/remote/usps_test.rb
shime: Wanted to follow the style of the entire project here, I can open up a separate PR which updates all hash styles.

lib/active_shipping/carriers/usps.rb
shime: I've basically mocked all the properties that are needed here - wanted to get some review before adding more stuff. Should I add the attributes needed here to `Package`?

test/remote/usps_test.rb
jonathankwok: Sounds good.

lib/active_shipping/carriers/usps.rb
jonathankwok: Also, why is it a 2D array when you use it like a Hash? You're using it like a [Hash here anyway, when you iterate over the key/value pairs](https://github.com/Shopify/active_shipping/pull/411/files#diff-8600a1e253812a15366c2ae13218aed6R460).

lib/active_shipping/carriers/usps.rb
jonathankwok: Yeah, we presently have the [class `PackageItem` which is perfect for a customs item within a package.](https://github.com/Shopify/active_shipping/blob/master/lib/active_shipping/package.rb#L5); however, it's not used in `active_shipping` explicitly. 

If you'd like to introduce a `PackageItem` array attribute on `Package`, I'd welcome that - alternatively, you could store it in the options during initialization, e.g. 

```
package_items = [pkg_item1, pkg_item2]
package = Package.new(10, [10, 20, 30], package_items: package_items)
```

This pattern is a little undesirable because it places something relatively important into an `options` Hash but it would work right out of the box.

lib/active_shipping/carriers/usps.rb
jonathankwok: That shouldn't be necessary - rather than assume it's `US` when no country is set, we should raise on a missing country in a `Location`; when it comes to the location elements of shipping, a country is one of the most important factors.

lib/active_shipping/carriers/usps.rb
shime: I don't follow. Looks like `ActiveShipping::Location#address1` is exactly what we need here for `Address1`, since `ActiveShipping::Location#address2` contains additional information about that address in cases you're referring to.

shime:
Regarding to making `country` mandatory for 'ActiveShipping::Location' - I would open a separate PR for that, since it's not related to USPS label retrieval.

lib/active_shipping/carriers/usps.rb
shime: Converted it to hash and added better naming.

lib/active_shipping/carriers/usps.rb
shime: I would open a separate PR for this.

lib/active_shipping/carriers/usps.rb
shime: How important is this? Can we rather do it in a separate PR?

shime:
What else is needed to get this merged into master?

lib/active_shipping/carriers/usps.rb
jonathankwok: This is pretty important - customs data should be accurate when creating a shipment. The contents and their weight/value can affect the final validity of the label request and more importantly, you can't physically ship something internationally if the customs data is missing.

It doesn't make sense to me to ship those in two separate PRs - you'd be shipping not just an incomplete, but misleading feature.

test/fixtures/xml/usps/domestic_label_response.xml
```diff
@@ -0,0 +1,550 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<DelivConfirmCertifyV4.0Response>
+  <DeliveryConfirmationNumber>420902109400101699320000720205</DeliveryConfirmationNumber>
+  <DeliveryConfirmationLabel>JVBERi0xLjINCjUgMCBvYmoNCjw8DQovVHlwZSAvWE9iamVjdA0KL1N1YnR5cGUgL0ltYWdlDQov
```
jonathankwok: You don't really need the binary representations of the label images in the test - feel free to replace these with some kind of placeholder/marker, like `<LABEL IMAGE REDACTED DUE TO SIZE>`

test/remote/usps_test.rb
```diff
@@ -125,6 +125,73 @@ def test_international_rates
     assert_equal [], other_than_two, "Some RateEstimates do not refer to the right number of packages (#{other_than_two.inspect})"
   end
 
+  def test_domestic_label_from_name_errros
```
jonathankwok: `errros` should be `errors`

test/fixtures/xml/usps/domestic_label_response.xml
```diff
@@ -0,0 +1,19 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<DelivConfirmCertifyV4.0Response>
+  <DeliveryConfirmationNumber><DELIVERY CONFIRMATION NUMBER></DeliveryConfirmationNumber>
```
jonathankwok: Ah, this is my fault - in the review I said you should use `<LABEL IMAGE REDACTED DUE TO SIZE>` but the angle brackets (`<`/`>`) would mess with the parsing. You can omit those. 😅

lib/active_shipping/carriers/usps.rb
shime: Thanks for the feedback, I've updated the code to use `PackageItem` and added some validations to the `ShippingContents` node.

lib/active_shipping/carriers/usps.rb
```diff
@@ -465,14 +468,39 @@ def build_world_shipment_request(action, origin, destination, package, options =
...
+
+            package.package_items.each do |item|
+              xml.ItemDetail do |xml|
+                if item.description.blank?
+                  raise PackageItemError.new("package must have a description")
+                else
+                  xml.Description(item.description)
+                end
+
+                xml.Quantity(1)
```
jonathankwok: You can pull in the `item.quantity` here: https://github.com/Shopify/active_shipping/blob/master/lib/active_shipping/package_item.rb#L16

lib/active_shipping/carriers/usps.rb
```diff
@@ -465,14 +468,39 @@ def build_world_shipment_request(action, origin, destination, package, options =
...
+
+                if item.value.blank?
+                  raise PackageItemError.new("package must have a value")
+                else
+                  xml.Value(item.value)
+                end
+
+                xml.NetPounds(0)
+                xml.NetOunces("%0.1f" % [package.ounces, 1].max)
```
jonathankwok: Should be `[package.ounces, 0].max`, it should be okay for the `NetOunces` to be 0.

lib/active_shipping/package_item.rb
```diff
@@ -17,6 +19,8 @@ def initialize(name, grams_or_ounces, value, quantity, options = {})
 
       @sku = options[:sku]
       @hs_code = options[:hs_code]
+      @description = options[:description]
+      @country_of_origin = options[:country_of_origin] || "US"
```
jonathankwok: Don't think it's really necessary to default a country of origin to US.

shime:
Thanks for the kind words, I've just made some more updates per your comments.

jimhj:
👍 
Any plan to merge this PR?

lib/active_shipping/location.rb
```diff
@@ -161,5 +164,17 @@ def address2_and_3
...
+    def first_name
+      @name.split(' ').first
+    end
+
+    def last_name
+      @name.split(' ').last
+    end
+
+    def domestic?
+      DOMESTIC_CODES.include?(country_code)
```
kmcphillips: This assumed domestic is american, which is an incorrect assumption.

lib/active_shipping/location.rb
```diff
@@ -161,5 +164,17 @@ def address2_and_3
     def ==(other)
       to_hash == other.to_hash
     end
+
+    def first_name
+      @name.split(' ').first
+    end
+
+    def last_name
+      @name.split(' ').last
```
kmcphillips: Do we not have any better way of doing this? This breaks in plenty of cases. `Von Bargen` `Van Gelder`

lib/active_shipping/location.rb
```diff
@@ -161,5 +164,17 @@ def address2_and_3
     def ==(other)
       to_hash == other.to_hash
     end
+
+    def first_name
+      @name.split(' ').first
```
kmcphillips: `Jane Willson Smith`, her middle name gets dropped too.

lib/active_shipping/location.rb
```diff
@@ -2,6 +2,9 @@ module ActiveShipping #:nodoc:
   class Location
     ADDRESS_TYPES = %w(residential commercial po_box)
 
+    US_POSSESSIONS = %w(AS FM GU MH MP PW PR VI)
+    DOMESTIC_CODES = US_POSSESSIONS + ['US', nil]
```
kmcphillips: These are USPS specific and should not be in the generic location

PR #413, opened 2016-11-14
[ci-skip] Add links to README.md

alexcameron89:
I added links to several of the files mentioned in the README so that it's easy to access the mentioned files. I also turned the plaintext URL's into links so that they can be clicked on the [shopify.github.io](http://shopify.github.io/active_shipping/) page.


kmcphillips:
Thanks!

PR #416, opened 2016-11-14
Add price_details attribute to RateEstimate

lucasuyezu:
### Problem
Currently there is no way to detect which services provide included features, like insurance and tracking. There is `shipment_options` but these are used for options that the merchant might select on top of the included ones.

### Solution
Add a new attribute `price_details` that can be initialized with detailed informations on each item of a rate estimate.

@kmcphillips @jonathankwok @mdking


mdking:
Actually sorry should add a test similar to other [options on the class](https://github.com/Shopify/active_shipping/blob/master/test/unit/rate_estimate_test.rb#L49-L64)

PR #420, opened 2016-11-24
Handle UPS tracking responses with no status code

jonathankwok:
## About
Some `TrackResponse`s from UPS don't have a status node, but are considered "Successful" responses. The current implementation expects all successful responses to have a status node but that's not true in all cases - the XML fixture I added is a real-world example of a response that has no status node but is considered a successful response.

This PR prevents the `status_code.at('Code').text` from failing on the `.text` call by not reading through the non-existent status nodes. 

What this means is that at the end when building `TrackingResponse`, the `@status` field will be empty; this is preferable to raising without returning anything so this is an acceptable side-effect.

@benwah @thegedge @iWuzHere

### DIFF
--- a/lib/active_shipping/carriers/ups.rb
+++ b/lib/active_shipping/carriers/ups.rb
@@ -816,16 +816,18 @@ module ActiveShipping
         # Build status hash
         status_nodes = first_package.css('Activity > Status > StatusType')
 
-        # Prefer a delivery node
-        status_node = status_nodes.detect { |x| x.at('Code').text == 'D' }
-        status_node ||= status_nodes.first
+        if status_nodes.present?
+          # Prefer a delivery node
+          status_node = status_nodes.detect { |x| x.at('Code').text == 'D' }
+          status_node ||= status_nodes.first
 
-        status_code = status_node.at('Code').text
-        status_description = status_node.at('Description').text
-        status = TRACKING_STATUS_CODES[status_code]
+          status_code = status_node.at('Code').try(:text)
+          status_description = status_node.at('Description').try(:text)
+          status = TRACKING_STATUS_CODES[status_code]
 
-        if status_description =~ /out.*delivery/i
-          status = :out_for_delivery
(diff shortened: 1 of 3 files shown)


lib/active_shipping/carriers/ups.rb
```diff
@@ -816,16 +816,18 @@ def parse_tracking_response(response, options = {})
         # Build status hash
         status_nodes = first_package.css('Activity > Status > StatusType')
 
-        # Prefer a delivery node
-        status_node = status_nodes.detect { |x| x.at('Code').text == 'D' }
-        status_node ||= status_nodes.first
+        if status_nodes.present?
```
thegedge: I think `status` can be undefined now (on line 838). We'll need to set a default.

jonathankwok: If defined within a conditional, it will be `nil`, not `undefined`, which means the check on line 838 will be `unless nil == :delivered`.

Is that acceptable, or should we still set some kind of status?

thegedge: Well, TIL. We should probably have a non-`nil` status (e.g., `:unknown`) when constructing the `TrackingResponse` further below

jonathankwok: We have an existing 💩  pattern of doing this, `nil` is a known possible result of a `status`. 

[USPS does the exact same thing](https://github.com/Shopify/active_shipping/blob/master/lib/active_shipping/carriers/usps.rb#L636)
[FedEx](https://github.com/Shopify/active_shipping/blob/master/lib/active_shipping/carriers/fedex.rb#L632) doesn't validate that the status code isn't `nil`.

thegedge: If we're already using `nil` as a status, then it's 👍 

We really should avoid `nil` in another PR though, and have an `:unknown` status so there's no room for interpretation of a `nil` status.

jonathankwok:
CI failing because it's Thanksgiving 😭

PR #421, opened 2016-11-28
Do not raise if FedEx tracking response is successful with no status

jonathankwok:
## Problem
We're pre-emptively raising when a `TrackResponse` from FedEx has no `StatusCode`, even [though their schema says `minOccurs = 0`](https://www.fedex.com/templates/components/apps/wpor/secure/downloads/pdf/201607/FedEx_WebServices_TrackService_WSDLGuide_v2016.pdf). It's a valid response, there's no reason to raise an `ActiveShipping::Error` if a response doesn't have a status. Rather, we should permit the return of a response + the XML, so the user of this data can at least work with the response from FedEx.

@iWuzHere @MalazAlamir @kmcphillips


lib/active_shipping/carriers/fedex.rb
```diff
@@ -619,20 +619,17 @@ def parse_tracking_response(response, options)
...
-        if status_detail.nil?
-          raise ActiveShipping::Error, "Tracking response does not contain status information"
-        end
-
-        status_code = status_detail.at('Code').try(:text)
-        if status_code.nil?
-          raise ActiveShipping::Error, "Tracking response does not contain status code"
-        end
+        if status_detail.blank?
+          status_code, status, status_description, delivery_signature = nil
```
jonathankwok: Ultimately not necessary since variables defined within a conditional are `nil`-initialized if the conditional wasn't true, but I thought it'd be nice to make it clear exactly what variables would be `nil` as a result of `status_detail` being blank.

PR #422, opened 2016-11-28
Sort FedEx tracking response error codes between retryable and not

jonathankwok:
Resource: https://www.fedex.com/us/developer/WebHelp/ws/2014/dvg/WS_DVG_WebHelp/index.htm#Appendix_O_Error_Code_Messages.htm

## Summary
Error codes are more extensive than known in FedEx, and for some errors, they were continuing past the error code check because they weren't known in the `case`. From the resource, I sorted the tracking errors (9xxx) between transient and non-transient, on the basis of whether the message contains "`Please retry later`" or not. 

This strategy aligns with what I've seen with existing numbers (e.g. 9080 is the error returned when I tried to track something with code `BadTrackingCode`, and the error response doesn't prompt a retry.)

As for the error classes being returned, I'm maintaining the current return pattern; `ShipmentNotFound` for transient errors, and `ResponseContentError` otherwise. Not great, but that's part of a larger scale rework coming later.

@iWuzHere @kmcphillips

### DIFF
--- a/lib/active_shipping/carriers/fedex.rb
+++ b/lib/active_shipping/carriers/fedex.rb
@@ -137,7 +137,11 @@ module ActiveShipping
     DEFAULT_LABEL_STOCK_TYPE = 'PAPER_7X4.75'
 
     # Available return formats for image data when creating labels
-    LABEL_FORMATS = ['DPL', 'EPL2', 'PDF', 'ZPLII', 'PNG'].freeze
+    LABEL_FORMATS = %w(DPL EPL2 PDF ZPLII PNG).freeze
+
+    TRANSIENT_TRACK_RESPONSE_CODES = %w(9035 9040 9041 9045 9050 9055 9060 9065 9070 9075 9085 9086 9090).freeze
+
+    UNRECOVERABLE_TRACK_RESPONSE_CODES = %w(9080 9081 9082 9095 9100).freeze
 
     def self.service_name_for_code(service_code)
       SERVICE_TYPES[service_code] || "FedEx #{service_code.titleize.sub(/Fedex /, '')}"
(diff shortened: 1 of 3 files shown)


lib/active_shipping/carriers/fedex.rb
```diff
@@ -137,7 +137,11 @@ class FedEx < Carrier
     DEFAULT_LABEL_STOCK_TYPE = 'PAPER_7X4.75'
 
     # Available return formats for image data when creating labels
-    LABEL_FORMATS = ['DPL', 'EPL2', 'PDF', 'ZPLII', 'PNG'] 
+    LABEL_FORMATS = %w(DPL EPL2 PDF ZPLII PNG).freeze
+
+    TRANSIENT_TRACK_RESPONSE_CODES = %w(9035 9040 9041 9045 9050 9055 9060 9065 9070 9075 9085 9086 9090).freeze
+
+    UNRECOVERABLE_TRACK_RESPONSE_CODES = %w(9080 9081 9082 9095 9100).freeze
```
kmcphillips: Intentionally unused?

jonathankwok: Yeah, at first I was thinking of doing three patterns, a `TRANSIENT`/`UNRECOVERABLE`/`UNKNOWN` but then I realized it doesn't really make sense to separate known unrecoverable with unknowns because it's really boils down to whether it's transient or not. 

Still good to have the constants though, in case in the future we want to discern between known unrecoverable failures and unknown unrecoverable errors.

lib/active_shipping/carriers/fedex.rb
```diff
@@ -603,18 +607,16 @@ def parse_tracking_response(response, options)
 
 
         first_notification = tracking_details.at('Notification')
-        if first_notification.at('Severity').text == 'ERROR'
-          case first_notification.at('Code').text
-          when '9040'
+        severity = first_notification.at('Severity').text
+        if severity == 'ERROR' || severity == 'FAILURE'
+          message = first_notification.try(:text)
```
MalazAlamir: `try!` here and below ?

jonathankwok: Can't `try!`, we still support [`active_support 3.2`](https://github.com/Shopify/active_shipping/blob/master/Gemfile.activesupport32), which does not have `try!` 

http://apidock.com/rails/v4.0.2/Object/try%21

PR #424, opened 2016-12-01
Run CI against rails master

kmcphillips:
@thegedge @jonathankwok @sgrif 

https://twitter.com/sgrif/status/804037955241312258

Let's run our CI against rails master.

The matrix is silly big right now. But we can be opinionated in the 2.0 branch and drop support for anything below 4.2.

### DIFF
--- a/Gemfile.activesupport_master
+++ b/Gemfile.activesupport_master
@@ -1,0 +1,4 @@
+source "https://rubygems.org"
+gemspec
+
+gem 'activesupport', github: 'rails/rails'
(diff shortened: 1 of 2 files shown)


sgrif:
Should master be added to allowed failures?

Gemfile.activesupport_master
```diff
@@ -0,0 +1,4 @@
+source "https://rubygems.org"
+gemspec
+
+gem 'activesupport', github: 'rails/rails', branch: 'master'
```
sgrif: No need to specify the branch

sgrif:
Might want to look at using appraisal for this, makes a lot of it much easier.

kmcphillips:
> Allowed failures

Probably not. This changes infrequently enough we should keep it working in all versions.

kmcphillips:
https://github.com/thoughtbot/appraisal Interesting. I'll cut an issue to do that as part of the release milestone.

PR #428, opened 2016-12-01
Added support for UPS Street Level Address Validation

kennon:
This adds support for the UPS street level address validation (XAV) endpoint. I tried to keep the interface as similar as possible to the existing stamps.com address validator, and I'm optimistic that this response interface could be used for FedEx as well, as it appears that the same kind of information is being returned in all cases.


lib/active_shipping/carriers/ups.rb
```diff
@@ -961,6 +977,95 @@ def parse_void_response(response, options={})
...
+
+      validity = nil
+      classification_code = nil
+      classification_description = nil
+      addresses = []
+
+      if success
+        classification_code = xml.at('AddressClassification/Code').text
+        
+        if classification_code == "1"
```
jonathankwok: A `case` statement would work here, e.g.

```ruby
classification = case classification_code
when "1"
  :commercial
when "2"
  :residential
else
  :unknown
end
```

lib/active_shipping/carriers/ups.rb
```diff
@@ -961,6 +977,95 @@ def parse_void_response(response, options={})
...
+          validity = :valid
+        elsif xml.at("AmbiguousAddressIndicator")
+          validity = :ambiguous
+        elsif xml.at("NoCandidatesIndicator")
+          validity = :invalid
+        else
+          validity = :unknown
+        end
+
+        xml.css('AddressKeyFormat').each do |address_key_node|
```
jonathankwok: For brevity, if you wish:
```ruby
addresses = xml.css('AddressKeyFormat`).map { |node| location_from_address_key_format_node(node) }
```

lib/active_shipping/carriers/ups.rb
```diff
@@ -961,6 +977,95 @@ def parse_void_response(response, options={})
...
+          addresses << location_from_address_key_format_node(address_key_node)
+        end
+      end
+
+      response = AddressValidationResponse.new(success, message, Hash.from_xml(response).values.first, :validity => validity, :classification => classification, :candidate_addresses => addresses, :xml => response, :request => last_request)
```
jonathankwok: Could you build out `params = Hash.from_xml(response).values.first` outside of the `AddressValidationResponse` construction? It's an added variable initialization but I feel like a little goes a long way in understanding exactly what you're trying to pass here, which is `params`.

test/unit/carriers/ups_test.rb
```diff
@@ -603,4 +603,13 @@ def test_package_surepost_less_than_one_lb_service_code
     assert_equal 'OZS', request.search('/Package/PackageWeight/UnitOfMeasurement/Code').text
     assert_equal '8.0', request.search('/Package/PackageWeight/Weight').text
   end
+
+  def test_address_validation
+    location = Location.new(address1: "55 Glenlake Parkway", city: "Atlanta", state: "GA", zip: "30328", country: "US")
+    address_validation_response = xml_fixture('ups/address_validation_response')
+    @carrier.expects(:commit).returns(address_validation_response)
+    response = @carrier.validate_address(location)
+    assert_equal :commercial, response.classification
```
jonathankwok: An added check for the `validity` seems like a good idea here. I mentioned it above, but I'd like to see some tests checking different validity responses, since right now we're only verifying the sunny day scenario.

kennon:
What exactly are you looking for in the tests? I couldn't determine based on the existing tests what all exactly needed to be tested. It is slightly silly doing a full iteration of tests all against an xml fixture...

As far as the documentation is concerned, id love to be able to link to their documentation, but I had to more or less guess based on sample PHP Code buried 6 layers deep in a zip file that you can only download after being logged in with a UPS account and navigating their completely schizophrenic developer website.

I looked at the other methods and none of them seemed to be particularly well documented. Anyhoo, please let me know what you are expecting and I'll add what I can. 👍

jonathankwok:
That's fair re: the lack of documentation, I know that feeling very well. 😛 

On the topic of testing, to clarify what I meant - it seems from your implementation that there are a variety of potential responses from UPS when validating addresses; you have a test that checks vs a `:valid` response, but what if the response was `:ambiguous` or `:invalid`? Adding fixtures and tests would be useful for several reasons - it'd of course make future contributors more confident when modifying code, as well as providing more insight into a relatively obfuscated API.

Does that make sense? You have `test_address_validation`, but I'd like to see `test_address_validation_for_valid`, `test_address_validation_for_ambiguous`,  and `test_address_validation_for_invalid`, at the least.

kennon:
@jonathankwok not sure if you got automatically notified or not, but I made your requested changes (y)

kennon:
@jonathankwok any chance of getting this merged?

jonathankwok:
Hey @kennon! Sorry about the delay - I'll merge this to `master`, and backport it to `1-9-stable`, our current stable branch. This is because we are currently working on a new major release of ActiveShipping and `master` will become unstable soon - I advise pointing your Gem version to that release after this is merged.

PR #432, opened 2016-12-03
Clean up test output and fix warnings

kmcphillips:
Test output is polluted with so many warnings.

kevin_kevify____source_active_shipping__zsh_

This reduces the extra verbose mode, and fixes a bunch of the ones that were still logged.

kevin_kevify____source_active_shipping__zsh_

### DIFF
--- a/lib/active_shipping/carriers/canada_post_pws.rb
+++ b/lib/active_shipping/carriers/canada_post_pws.rb
@@ -1,6 +1,6 @@
 module ActiveShipping
   class CanadaPostPWS < Carrier
-    
+
     cattr_reader :name
     @@name = "Canada Post PWS"
 
@@ -569,13 +569,13 @@ module ActiveShipping
 
     def tracking_url(pin)
       case pin.length
-        when 12, 13, 16
-          endpoint + "vis/track/pin/%s/detail" % pin
-        when 15
-          endpoint + "vis/track/dnc/%s/detail" % pin
-        else
-          raise InvalidPinFormatError
-        end
+      when 12, 13, 16
+        "#{endpoint}vis/track/pin/#{pin}/detail"
+      when 15
+        "#{endpoint}vis/track/dnc/#{pin}/detail"
+      else
+        raise InvalidPinFormatError
+      end
     end
 
     def create_shipment_url(options)
--- a/test/remote/fedex_test.rb
+++ b/test/remote/fedex_test.rb
@@ -225,7 +225,7 @@ class RemoteFedExTest < Minitest::Test
     assert_equal "Delivered", response.status_description
 
     assert_equal Time.parse('Fri, 03 Jan 2014'), response.ship_time
-    assert_equal nil, response.scheduled_delivery_date
+    assert_nil response.scheduled_delivery_date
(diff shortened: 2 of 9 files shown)


kmcphillips:

lib/active_shipping/carriers/canada_post_pws.rb
```diff
@@ -569,13 +569,13 @@ def log(msg)
...
-        else
-          raise InvalidPinFormatError
-        end
+      when 12, 13, 16
+        endpoint + "vis/track/pin/%s/detail" % pin
+      when 15
+        endpoint + "vis/track/dnc/%s/detail" % pin
+      else
+        raise InvalidPinFormatError
+      end
```
thegedge: Up to you, but while you're here, maybe would be nice to switch to string interpolation:
```ruby
case pin.length
when 12, 13, 16
  "#{endpoint}/vis/track/pin/#{pin}/detail"
when 15
  "#{endpoint}/vis/track/dnc/#{pin}/detail"
else
  raise InvalidPinFormatError
end
```

test/remote/fedex_test.rb
```diff
@@ -391,7 +391,7 @@ def test_obtain_shipping_label_with_label_format_option
 
     assert response.success?
     refute_empty response.labels
-    data = response.labels.first.img_data 
+    data = response.labels.first.img_data
     refute_empty data
     assert data[0...4] == '%PDF'
```
thegedge: Feel free to keep a note for later, but given that `nil` doesn't have `empty?`, I think we could simplify all of this to `assert_equal '%PDF', response.labels.first.img_data[0...4]`

test/unit/external_return_label_request_test.rb
```diff
@@ -118,24 +118,24 @@ def test_customer_address2
...
 
   def test_sanitize
-    assert_equal @external_request_label_req.send(:sanitize,'   ').length, 0
-    assert_equal @external_request_label_req.send(:sanitize, 'some string   '), 'some string'
-    assert_equal @external_request_label_req.send(:sanitize, {}), nil
-    assert_equal @external_request_label_req.send(:sanitize, nil), nil
-    assert_equal @external_request_label_req.send(:sanitize, nil), nil
-    assert_equal @external_request_label_req.send(:sanitize, []), nil
-    assert_equal @external_request_label_req.send(:sanitize, (1..100).to_a.join("_")).size, ExternalReturnLabelRequest::CAP_STRING_LEN
+    assert_equal 0, @external_request_label_req.send(:sanitize,'   ').length
```
thegedge: `assert_equal "", @external_request_label_req.send(:sanitize,'   ')` is definitely a better option here.

kmcphillips:
I'm trying to do the minimum functional changes as possible here. So I didn't update all the interpolation in the file because _technically_ it's a behaviour change:

```
[1] pry(main)> nil + "asdf"
NoMethodError: undefined method `+' for nil:NilClass
from (pry):1:in `__pry__'
[2] pry(main)> "#{nil}asdf"
=> "asdf"
```

But you're right, and did update the one that I touched.

thegedge:
Good point! We don't have `nil` concerns here, but that's good thing for me to keep in mind in the future.

PR #433, opened 2016-12-03
Fix UPS remote tests on weekends

kmcphillips:
UPS remote test fails on the weekend.

One was not taking weekends into account at all. The other appears to be an exception around next day showing up on the Monday, not the Tuesday. Less confident about that one, but can't test it until the week anyway.


kmcphillips:
Simplified this a bit. One change immediately failed. So this fixes the one case that doesn't take workdays into account at all.

Issue #434, opened 2016-12-05
Improve clarity/usage of Package dimensions


jonathankwok:
## Problem
If you weren't aware of the implementation, what do you think are the details of this package?

```ruby
ActiveShipping::Package.new(20, [13, 14, 15])
```

What is the `20`? 
What is the dimensional ordering of `13x14x15`? Hint: It's not LWH.

Even if you add `units: :imperial`, it doesn't explain what units you're using, just the system the units would be in. Is it inches? Feet? Furlongs? Fathoms? 

Also, retrieving the length is not as easy as you would think.

```
irb(main):001:0> package = ActiveShipping::Package.new(20, [13, 14, 15])
=> #<ActiveShipping::Package:0x007f957e15ecd8 @options={}, @dimensions=[#<Quantified::Length: 13 centimetres>, #<Quantified::Length: 14 centimetres>, #<Quantified::Length: 15 centimetres>], @weight_unit_system=:metric, @dimensions_unit_system=:metric, @weight=#<Quantified::Mass: 20 grams>, @value=nil, @currency=nil, @cylinder=false, @gift=false, @oversized=false, @unpackaged=false>
irb(main):002:0> package.length
NoMethodError: undefined method `length' for #<ActiveShipping::Package:0x007f957e15ecd8>
```

No, you would have to do this:
```
irb(main):009:0> package.centimetres[2]
=> 15
```

Or is it `[0]`? Who can remember. Why isn't it `package.length`? Why does it return a `Fixnum` and not a `Quantified`? Why are we using Quantified?

## Possible solution
Now, what if we changed it to be more verbose?

```ruby
ActiveShipping::Package.new(
  weight: Measured::Weight.new("20", "g"),
  length: Measured::Length.new("15", "g"),
  width: Measured::Length.new("14", "g"),
  height: Measured::Length.new("13", "g"),
) 
```

Zero chance of confusion. In addition, uses the [`measured`](https://github.com/Shopify/measured/) gem, which is what we want in #427. It's more lines/typing, but it eliminates all areas of uncertainty. And when you're shipping something, certainty is key.

CC: @kmcphillips @thegedge

thegedge:
I think the use of measured is great here. And, if desired, I'm going to add an extension module to number classes (opt-in) that will make this far less verbose (e.g., `15.grams`).

jonathankwok:
Absolutely, I'm on board with a less verbose version, but not before having something clear. An express way of implementing the above could be:

``` ruby
[...]

Issue #427, opened 2016-12-01
Replace Quantified with Measured


kmcphillips:
We have put a ton of work into [measured](https://github.com/Shopify/measured) and are using it throughout our projects at Shopify. 

[In the README](https://github.com/Shopify/measured#gem-quantified) we discuss all the reasons that `measured` provides more features and stability and reliability over `quantified`. So let's swap them out.

@thegedge @jonathankwok

benwah:
Done here: https://github.com/Shopify/active_shipping/pull/450

PR #435, opened 2016-12-06
Split up and test Package and PackageItem

kmcphillips:
# What's the deal with `Package` and `PackageItem`

## Problem

These classes are a bit of a mess. The APIs are not clear, they don't do what you think they'd do, they have implicit behaviour, they're mostly untested, and we have no confidence in changing them.

## Solution

Split the class out into its own file. Test the specs as they exist.

There are even a bunch of cases where calling them with valid APIs have bugs and they raise. I'm not changing _any_ behaviour. Just asserting the raises and skipping the tests. If the tests stop raising they'll then fail.

This lets us pull out `Quantified` and replace with `measured` as in #434. This is really just a first step in that refactor.

Note: I wrote so many tests you may have to click `View Diff` because the last file with 400+ lines is collapsed.


test/unit/package_test.rb
```diff
@@ -1,68 +1,403 @@
...
-                                    :currency => 'USD'
-                        )
-
-    @metric_package = Package.new(100, [5, 18.5, 40],
-                                  :value => 860,
-                                  :currency => 'CAD'
-                      )
-  end
-
-  def test_accessors
```
kmcphillips: This test was incomplete and wasn't really testing everything clearly. Replaced it with many other tests testing each feature explicitly.

lib/active_shipping/package_item.rb
```diff
@@ -0,0 +1,72 @@
+module ActiveShipping #:nodoc:
+  class PackageItem
```
kmcphillips: Nothing in here changed. Just moved.

kmcphillips:
A first step toward #427 in 2.0

test/unit/package_item_test.rb
```diff
@@ -0,0 +1,237 @@
...
+      assert_instance_of ::Quantified::Mass, @item.weight(type: :billable)
+      flunk "This code path is broken but passed unexpectedly"
+    rescue NoMethodError
+      skip "This code path is broken"
+    end
+  end
+
+  def test_grams_value
+    assert_equal 100, @item.grams
+  end
```
thegedge: This seems inconsistent. We have `test_initialize_with_all_attributes` and then this one. We're testing the same behaviour (state on initialization), but across two tests.

This is likely a difference between classical unit testing and behaviour-driven testing, where the former focuses more on individual methods, but I'm more for the latter. What are your thoughts?

test/unit/package_item_test.rb
```diff
@@ -0,0 +1,237 @@
...
+      flunk "This code path is broken but passed unexpectedly"
+    rescue NoMethodError
+      skip "This code path is broken"
+    end
+  end
+
+  def test_ounces_converts
+    @item = PackageItem.new(@name, @weight, @value, @quantity, @options.merge(units: :metric))
+
+    assert_in_delta 3.5, @item.ounces, 1
```
thegedge: Again, large delta.

test/unit/package_item_test.rb
```diff
@@ -0,0 +1,237 @@
...
+      flunk "This code path is broken but passed unexpectedly"
+    rescue
+      skip "This code path is broken"
+    end
+  end
+
+  def test_pounds_converts
+    @item = PackageItem.new(@name, 1000, @value, @quantity, @options.merge(units: :metric))
+
+    assert_in_delta 2.2, @item.pounds, 1
```
thegedge: I'll just let you find all of these :)

test/unit/package_item_test.rb
```diff
@@ -0,0 +1,237 @@
...
+    assert_equal 100, @item.grams
+  end
+
+  def test_grams_accepts_options_with_type
+    begin
+      assert_equal :todo, @item.grams(type: :volumetric)
+      flunk "This code path is broken but passed unexpectedly"
+    rescue NoMethodError
+      skip "This code path is broken"
+    end
```
thegedge: A couple of things:

1. Although this is slightly different, I wonder would `assert_raises(NameError) { ... }` basically do the same thing for us?
2. If we want this exactly, it's used enough that maybe a helper function in `TestHelper` would be nice? `broken_with(NameError) { ... }`, or whatever name you please. I'm fine keeping these here too, since they'll hopefully disappear soon!

test/unit/package_item_test.rb
```diff
@@ -0,0 +1,237 @@
...
+  end
+
+  def test_grams_converts
+    @item = PackageItem.new(@name, 100, @value, @quantity, @options.merge(units: :imperial))
+
+    assert_in_delta 2834.9, @item.grams, 1
```
thegedge: 1 is a very large window of error, which makes me lose trust in the library. I'd tighten up the error bounds, e.g., `assert_in_delta 2834.95, @item.grams, 0.01`.

test/unit/package_test.rb
```diff
@@ -1,68 +1,403 @@
...
+  end
+
+  def test_initialize_weight_units
+    @package = Package.new(@weight, @dimensions, @options.merge(weight_units: :imperial))
+
+    assert_equal @weight, @package.ounces
+  end
+
+  def test_unpackaged
+    assert Package.new(@weight, @dimensions, unpackaged: true).unpackaged?
```
thegedge: Always avoid bare `assert` / `refute` (bare meaning without an error message) since the failed assertion message is useless. `assert_predicate Package.new(...), :unpackaged?` for this, and anywhere else here.

test/unit/package_test.rb
```diff
@@ -1,68 +1,403 @@
...
+    assert_equal @dimensions[1], @imperial_package.inches(:wide)
+  end
+
+  def test_inches_z
+    assert_equal @dimensions[0], @imperial_package.inches(:z)
+    assert_equal @dimensions[0], @imperial_package.inches(:min)
+    assert_equal @dimensions[0], @imperial_package.inches(:height)
+    assert_equal @dimensions[0], @imperial_package.inches(:depth)
+    assert_equal @dimensions[0], @imperial_package.inches(:high)
+    assert_equal @dimensions[0], @imperial_package.inches(:deep)
```
thegedge: Just for noting: I'm personally not a fan of this behaviour. It's adding unnecessary complexity / surface area to this library. Are they useful? My own opinion is that they aren't, and that we should let clients decide if they want to alias things.

test/unit/package_test.rb
```diff
@@ -1,68 +1,403 @@
...
+
+  def test_cents_from_string
+    assert_equal 120, Package.cents_from("1.20")
+  end
+
+  def test_cents_from_int
+    assert_equal 12, Package.cents_from(12)
+  end
+
+  def test_cents_from_nonsense
```
thegedge: So much 😂 for this name. I think I'd be on the floor if it were `noncents`...

test/unit/package_test.rb
```diff
@@ -1,68 +1,403 @@
...
+    assert_equal @dimensions[0], @imperial_package.inches(0)
+    assert_equal @dimensions[1], @imperial_package.inches(1)
+    assert_equal @dimensions[2], @imperial_package.inches(2)
+    assert_nil @imperial_package.inches(3)
+  end
+
+  def test_inches_x
+    assert_equal @dimensions[2], @imperial_package.inches(:x)
+    assert_equal @dimensions[2], @imperial_package.inches(:max)
+    assert_equal @dimensions[2], @imperial_package.inches(:length)
```
jonathankwok: I had _no idea_ this was possible.

test/unit/package_test.rb
```diff
@@ -1,68 +1,403 @@
...
+  end
+
+  def test_centimetres_girth_cylinder
+    @package = Package.new(@weight, @dimensions, @options.merge(cylinder: true, units: :metric, dim_units: :metric))
+
+    assert @package.cylinder?
+    assert_in_delta 17.2, @package.centimetres(:girth), 1
+    assert_in_delta 17.2, @package.centimetres(:around), 1
+    assert_in_delta 17.2, @package.centimetres(:circumference), 1
+
```
jonathankwok: Extra 🌠

test/unit/package_test.rb
kmcphillips: Let's make it impossible ok? And make a more normal looking API.

test/unit/package_item_test.rb
```diff
@@ -0,0 +1,237 @@
...
+    }
+    @item = PackageItem.new(@name, @weight, @value, @quantity, options)
+
+    assert_nil @item.hs_code
+    assert_nil @item.sku
+    refute_equal @weight, @item.ounces
+    assert_equal @weight, @item.grams
+  end
+
+  def test_initialize_negative_quantity
```
kmcphillips: I really dislike the `def test_` DSL here. It's part of what makes minitest so hard to work with.

I'll be coming back and naming all the tests with `test "#method_name action conditions" do` as part of 2.0. So I'm not really worried about the naming here.

test/unit/package_item_test.rb
```diff
@@ -0,0 +1,237 @@
...
+  end
+
+  def test_initialize_weight_accepts_metric
+    @item = PackageItem.new(@name, @weight, @value, @quantity, @options.merge(units: :metric))
+
+    assert_equal @weight, @item.grams
+    refute_equal @weight, @item.ounces
+  end
+
+  def test_initialize_weight_does_not_accept_strings
```
kmcphillips: All these tests test inconsistent behaviour. We are in agreement.

test/unit/package_item_test.rb
kmcphillips: Again. All part of a bad interface. In that the `grams` method is both an accessor and performing logic. There is no way to test these pieces in isolation. So I have opted to be a bit redundant, and test what is available, and all edges of the public interface. However awkward.

test/unit/package_item_test.rb
kmcphillips: We decided against 1. But 2, that's what this does. It doesn't do the indirection through a helper, but it definitely is explicit in that we expect a failure, then skip the test.

test/unit/package_item_test.rb
thegedge: I don't think this is part of a bad interface, but simply a mixture of testing paradigms. There's a paradigm where you focus on individual methods and the inputs/outputs of those. Then there's behavioural testing where you have expected behaviours of your classes and test those, which often means testing across methods/accessors/readers/etc. I'm a big fan of the latter, but it's fine to have some of the former too.

What's confusing to me, and the reason for this comment, is that we have `test_initialize_with_all_attributes` test, which tests the state of things after initialization, and that's exactly what this test is also doing. We've taken a behaviour (state immediately after initialization) and split it across the wrong boundary (`attr_reader` / `def foo`). Whether there's logic or not happening in those should be irrelevant to clients of the object. Also, there's nothing stopping us from "upgrading" one of the `attr_reader`s to a manual `def attr; ...stuff...; end` in the future.

Sorry for the wordiness, but hopefully that cleared up my stance :)

thegedge: P.S. I don't want to block on shipping because of this, but I do believe this would make for a higher quality and easier to understand test suite.

PR #436, opened 2016-12-07
Add 1.x vs 2.x notes to README

kmcphillips:
This is the last change I am going to introduce to master before branching.

I'll merge this and ship version `1.9` and create the branch `1-9-stable`.

Then onward and upward to `2.0`. I've got a bunch of changes on branches.

@mdking @jonathankwok @thegedge

### DIFF
--- a/README.md
+++ b/README.md
@@ -23,6 +23,14 @@ Active Shipping is currently being used and improved in a production environment
 * [Kunaki](http://www.kunaki.com)
 * [Australia Post](http://auspost.com.au/)
 
+
+## Versions
+
+Development on `master` is for a pre-release of the `2.x` version of this gem. It includes a number of backwards incompatible changes and improvements. Progress can be tracked in [the milestone](https://github.com/Shopify/active_shipping/milestone/1).
+
+We are no longer actively developing version `1.x`. We have released `1.9` and will only backport small fixes to this version, on branch `1-9-stable`, and they should be on `master` first.
+
+
 ## Installation
 
 Using bundler, add to the `Gemfile`:


README.md
```diff
@@ -23,6 +23,14 @@ Active Shipping is currently being used and improved in a production environment
 * [Kunaki](http://www.kunaki.com)
 * [Australia Post](http://auspost.com.au/)
 
+
+## Versions
+
+Development on `master` is for a pre-release of the `2.x` version of this gem. It includes a number of backwards incompatible changes and improvements. Progress can be tracked in [the milestone](https://github.com/Shopify/active_shipping/milestone/1).
+
+We are no longer actively developing version `1.x`. We have released `1.9` and will only backport small fixes to this version, on branch `1-9-stable`, and they should be on `master` first.
```
mdking: Should we instead keep `master` as stable release and create a `2-0` branch? This way development on `2-0` can proceed without needing to be functionally complete? Thoughts?

kmcphillips: I'm strongly opposed to this.

We want to encourage contributions, by us and by the OSS community, against the newest version. By default, look forward. We won't be unstable for long. And we can backport to `1-9-stable` in the cases where something is urgent.

This is also the way rails does it.

mdking: > We want to encourage contributions, by us and by the OSS community, against the newest version. 

I don't think either approach discourages OSS community contribution but agree it's easier for PRs as they  default to being raised against `master` than a specific version branch. 

I don't feel strongly about it as the readme does a good job outlining our intentions and branching strategy and that's what is key here 👍

PR #437, opened 2016-12-07
Add messages to a rate estimate

tamcgoey:
**Problem:**

Right now UPS will return a successful rate request but will pass along warning messages indicating that the request wasn't actually successful given the account status.

We're wanting to pass these warning messages along so that we can do things like notifying the client that they don't have negotiated rates despite rates returning successfully.

**Solution**

- Added a `:messages` field to `RateEstimate`. 
- Added `rate_warning_messages` to UPS to parse warning messages like negotiated rates not available to this specific account.

Review: @kmcphillips @jonathankwok 

cc @RichardBlair


lib/active_shipping/carriers/ups.rb
```diff
@@ -998,6 +1000,10 @@ def response_message(document)
       [status, desc].select(&:present?).join(": ").presence || "UPS could not process the request."
     end
 
+    def rate_warning_messages(rate_xml)
+      rate_xml.xpath("RatedShipmentWarning").map { |warning| warning.text }
```
kmcphillips: Will this return an empty set or `nil` if no elements match? Do we have an existing test that covers this?

tamcgoey: It will return an empty set. I've gone ahead and added a new test case to cover this.

kmcphillips: Great 👌

kmcphillips:
We'll release this in 1.9.1

PR #442, opened 2016-12-08
Use ActiveSupport::TestCase globally

kmcphillips:
No need to use the `Minitest` parent class, or the version swapping check. We've dropped lower versions. And can use `ActiveSupport::TestCase` and all the things it provides everywhere.

### DIFF
--- a/test/test_helper.rb
+++ b/test/test_helper.rb
@@ -13,14 +13,7 @@ require 'pry'
 
 Minitest::Reporters.use!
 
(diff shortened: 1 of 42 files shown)


test/test_helper.rb
```diff
@@ -13,14 +13,7 @@
 
 Minitest::Reporters.use!
 
-# This makes sure that Minitest::Test exists when an older version of Minitest
```
kmcphillips: This is the interesting bit

PR #443, opened 2016-12-08
Modernize tests - Part 1

kmcphillips:
Now that #442 is in I'm going to slowly start converting the fragile, confusing, and unclear `def test_` DSL and replace it with the `ActiveSupport::TestCase` `test()` block based DSL.

Changes are just around naming and formatting mostly. Split up test cases that do more than one thing onto distinct tests with a clearer fail message. Extracted some shared variables. In one case I replaced `assert_silent` with `assert_nothing_raised` because the original author I don't think understood the assertions. The former is about printing to standard out, but I'm sure they intended to deal with exceptions.

This does `Carrier` `Carriers` and `ExternalReturnLabelRequest`.

### DIFF
--- a/test/unit/external_return_label_request_test.rb
+++ b/test/unit/external_return_label_request_test.rb
@@ -3,121 +3,164 @@ require 'test_helper'
 class ExternalReturnLabelRequestTest < ActiveSupport::TestCase
   include ActiveShipping::Test::Fixtures
 
-  def setup
+  setup do
     @external_request_label_req =
-      ExternalReturnLabelRequest.from_hash ({
-        :customer_name => "Test Customer",
-        :customer_address1 => "122 Hudson St.",
-        :customer_city => "New York",
-        :customer_state => "NY",
-        :customer_zipcode => "10013",
-        :label_format => "No Instructions",
-        :label_definition => "4X6",
-        :service_type_code => "044",
-        :merchant_account_id => "12345",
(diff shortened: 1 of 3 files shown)


kmcphillips:
```
Searching 78 files for "def test_"

614 matches across 38 files
```

Keep on keepin on.

test/unit/external_return_label_request_test.rb
```diff
@@ -3,121 +3,164 @@
...
     assert_raises(USPSValidationError) do
-      @external_request_label_req.recipient_bcc = "not_a_valid_email"
+      @external_request_label_req.recipient_bcc = @invalid_email
     end
-    assert_silent do
-      @external_request_label_req.recipient_bcc = "no-reply@example.com"
+  end
+
+  test "#recipient_bcc assigns the email" do
+    assert_nothing_raised do
```
jonathankwok: [`assert_nothing_raised` is pretty useless at best and harmful at worst](http://www.thagomizer.com/blog/2013/06/11/assert-nothing-raised.html), since all Minitest test cases are expected to not raise.

kmcphillips: Well. I don't have a strong opinion. It's explicit in what it's asking for.

The major down side is then you have tests that have no explicit assertions. Only implicit expectations that nothing raises.

PR #446, opened 2016-12-13
Clean up tests and formatting for Location

kmcphillips:
All cleanup in `Location`.

No functional change. Cleans up formatting, extracts a constant, removes old hash syntax, expands methods to be more consistent.

Reformats the tests to be `test "" do` and adds a bunch of missed coverage.

### DIFF
--- a/lib/active_shipping/location.rb
+++ b/lib/active_shipping/location.rb
@@ -75,7 +75,7 @@ module ActiveShipping #:nodoc:
         aliases.detect do |sym|
           value = if hash_access
             object[sym]
-          elsif object.respond_to?(sym) && !Hash.public_instance_methods.include?(sym.to_s)
+          elsif object.respond_to?(sym)
             object.send(sym)
           end
 
--- a/lib/active_shipping/location.rb
+++ b/lib/active_shipping/location.rb
@@ -2,7 +2,7 @@ module ActiveShipping #:nodoc:
   class Location
     ADDRESS_TYPES = %w(residential commercial po_box)
 
-    ATTRIBUTE_MAPPINGS = {
+    ATTRIBUTE_ALIASES = {
       name: [:name],
       country: [:country_code, :country],
       postal_code: [:postal_code, :zip, :postal],
@@ -71,12 +71,15 @@ module ActiveShipping #:nodoc:
         false
       end
 
-      ATTRIBUTE_MAPPINGS.each do |pair|
-        pair[1].each do |sym|
-          if value = (object[sym] if hash_access) || (object.send(sym) if object.respond_to?(sym) && (!hash_access || !Hash.public_instance_methods.include?(sym.to_s)))
-            attributes[pair[0]] = value
-            break
+      ATTRIBUTE_ALIASES.each do |attribute, aliases|
+        aliases.detect do |sym|
+          value = if hash_access
+            object[sym]
+          elsif object.respond_to?(sym) && !Hash.public_instance_methods.include?(sym.to_s)
+            object.send(sym)
           end
+
+          attributes[attribute] = value if value
         end
       end
 
--- a/lib/active_shipping/location.rb
+++ b/lib/active_shipping/location.rb
@@ -2,6 +2,21 @@ module ActiveShipping #:nodoc:
   class Location
     ADDRESS_TYPES = %w(residential commercial po_box)
 
+    ATTRIBUTE_MAPPINGS = {
+      name: [:name],
+      country: [:country_code, :country],
+      postal_code: [:postal_code, :zip, :postal],
+      province: [:province_code, :state_code, :territory_code, :region_code, :province, :state, :territory, :region],
+      city: [:city, :town],
+      address1: [:address1, :address, :street],
+      address2: [:address2],
+      address3: [:address3],
+      phone: [:phone, :phone_number],
+      fax: [:fax, :fax_number],
+      address_type: [:address_type],
+      company_name: [:company, :company_name],
+    }.freeze
+
     attr_reader :options,
                 :country,
                 :postal_code,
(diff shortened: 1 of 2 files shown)


lib/active_shipping/location.rb
```diff
@@ -42,48 +60,50 @@ def initialize(options = {})
...
+
       hash_access = begin
         object[:some_symbol]
         true
       rescue
         false
       end
-      attr_mappings.each do |pair|
+
+      ATTRIBUTE_MAPPINGS.each do |pair|
```
jonathankwok: While we're here, we can make this a little clearer to read too. What do you think about these changes?

 - Renaming `ATTRIBUTE_MAPPINGS` to `ATTRIBUTE_ALIASES`
 - Change the `each` `pair` parameter to an actual pair of named parameters, `attribute` and `aliases`:
 - Change the `pair[1].each` to be a `detect` since that's what we're really doing anyway with the `break`.

So something like:
```ruby
ATTRIBUTE_ALIASES.each do |attribute, aliases|
  aliases.detect do |alias|
   # blah blah
```
It makes it abundantly clear that we're assigning the aliased attribute's value to whichever aliased attribute what passed in for that attribute.

lib/active_shipping/location.rb
```diff
@@ -129,9 +149,9 @@ def inspect
     # Returns the postal code as a properly formatted Zip+4 code, e.g. "77095-2233"
     def zip_plus_4
       if /(\d{5})(\d{4})/ =~ @postal_code
-        return "#{$1}-#{$2}"
```
jonathankwok: We can simplify this, I believe;

```ruby
def zip_plus_4
  "#{$1}-#{2} if /(\d{5})-?(\d{4})/ =~ @postal_code
end
```

Tested it out a bit:

```ruby
irb(main):001:0> "#{$1}-#{$2}" if /(\d{5})-?(\d{4})/ =~ "12345-1234"
=> "12345-1234"
irb(main):002:0> "#{$1}-#{$2}" if /(\d{5})-?(\d{4})/ =~ "123451234"
=> "12345-1234"
irb(main):003:0> "#{$1}-#{$2}" if /(\d{5})-?(\d{4})/ =~ "K2P 1L4"
=> nil
```

test/unit/location_test.rb
```diff
@@ -28,108 +36,156 @@ def test_location_from_strange_hash
...
+  test "#inspect returns a readable string" do
+    expected = "110 Laurier Avenue West\nOttawa, ON, K1P 1J1\nCanada\nPhone: 1-613-580-2400\nFax: 1-613-580-2495"
+    assert_equal expected, @location.inspect
   end
 
-  def test_set_address_type
-    location = location_fixtures[:ottawa]
-    assert !location.commercial?
+  test "#address_type= assigns a type of address as commercial" do
+    refute @location.commercial?
```
thegedge: We should switch to `assert_predicate` here and everywhere else. Okay to leave this for a future PR, but best to create an issue if deferring. Feel free to assign to me.

lib/active_shipping/location.rb
kmcphillips: I just noticed the `if value = (object[sym] if hash_access)` gross.

jonathankwok: Yeah, mentioned it offline but `object.send(sym) if object.respond_to?(sym)` should be synonymous to `object.try(sym)` in this case.

test/unit/location_test.rb
kmcphillips: #447

kmcphillips:
Reping with changes.

test/unit/location_test.rb
```diff
@@ -3,133 +3,214 @@
...
+    assert_equal @attributes_hash[:territory_code], location.province
+    assert_equal @attributes_hash[:town], location.city
+    assert_equal @attributes_hash[:address], location.address1
+    assert_equal @attributes_hash[:phone], location.phone
+    assert_equal @attributes_hash[:fax_number], location.fax
+    assert_equal @attributes_hash[:address_type].to_s, location.address_type
+    assert_equal @attributes_hash[:name], location.name
+  end
+
+  test ".from sets from an object with properties" do
```
kmcphillips: We didn't actually have any tests for `.from` an object, not a hash.

This `Class.new` is a little nasty, but I tried with `OpenStruct` and it still responds to `:[]` so doesn't work.

lib/active_shipping/location.rb
```diff
@@ -42,48 +60,53 @@ def initialize(options = {})
...
         false
       end
-      attr_mappings.each do |pair|
-        pair[1].each do |sym|
-          if value = (object[sym] if hash_access) || (object.send(sym) if object.respond_to?(sym) && (!hash_access || !Hash.public_instance_methods.include?(sym.to_s)))
-            attributes[pair[0]] = value
-            break
+
+      ATTRIBUTE_ALIASES.each do |attribute, aliases|
+        aliases.detect do |sym|
```
kmcphillips: This is a _little_ weird in that `detect` we don't care about the return value, but we do want it to exit early. Test fails if you change it to `each`.

lib/active_shipping/location.rb
```diff
@@ -42,48 +60,53 @@ def initialize(options = {})
...
+          value = if hash_access
+            object[sym]
+          elsif object.respond_to?(sym) && !Hash.public_instance_methods.include?(sym.to_s)
```
jonathankwok: I'm trying to understand what the second check there is for. Two questions:

1. `.public_instance_methods` returns an array of symbols, so the `include` would always be `false` and result in this predicate always being `true`.
2. "Call this method only if Hash objects don't have it"? What does that give us? None of the aliased symbols are instance methods on Hash.

kmcphillips: Yes. You're of course right.

kmcphillips: Removed

kmcphillips:
Good @jonathankwok ?

kmcphillips:
Oops, didn't squash because I'm a chuuuuuuump... :persevere:

jonathankwok:
At least they're not goofy commit names 😛

PR #451, opened 2016-12-21
#date_for should return a Date, not a DateTime

dustMason:
I believe this method is the source of a tricky bug in my application, and I suspect it is in others as well.

I discovered it while inspecting API responses from Stamps.com. As indicated by the method signature, API responses are expected to provide _calendar_ date information – any timezone information is discarded here, which is correct. However, the `#date_for` method returns an instance of DateTime, which includes time zone information. In this case that time zone is set to the default value of UTC (+0000) which is likely not correct.

If the resulting value is passed, for instance, to a Rails app that saves it in a timestamp field, it will be erroneously converted to a local time thus possibly changing the date (as it does in my application).

Changing this method to return a Date instance (as documented in the comment above `#date_for`) prevents the extra timezone information from being added by the DateTime constructor, which we don't want because the `#strptime` arguments cause them to be lost.

### DIFF
--- a/lib/active_shipping/rate_estimate.rb
+++ b/lib/active_shipping/rate_estimate.rb
@@ -189,7 +189,7 @@ module ActiveShipping
     # @return [Date, nil] The Date object absed on the input, or `nil` if no date
     #   could be determined.
     def date_for(date)
-      date && DateTime.strptime(date.to_s, "%Y-%m-%d")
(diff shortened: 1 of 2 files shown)


dustMason:
I would be happy to fix up the tests and amend the PR.

dustMason:
I changed the assertions to expect Date vs DateTime. Let me know if there's anything else needed.

PR #452, opened 2016-12-21
Added support for Holiday customization in tests

benwah:
This allows us to wrap assertions around `with_holidays(:carrier)` - And lets us support other carriers such as usps, fedex in the future.

FedEX will be more complicated to support: http://www.fedex.com/us/service-guide/holiday-schedule.html
USPS pretty simple: https://about.usps.com/news/events-calendar/2016-postal-holidays.htm

### DIFF
--- a/test/helpers/holiday_helpers.rb
+++ b/test/helpers/holiday_helpers.rb
@@ -1,0 +1,46 @@
+module HolidayHelpers
+  class MissingConfigurationError < RuntimeError; end
+
+  HOLIDAYS = {
+    ups: { # https://compass.ups.com/ups-holiday-schedule-2016/
+      "2016" => [
+        { month: 1,  day: 1  },
+        { month: 5,  day: 30 },
+        { month: 6,  day: 4  },
+        { month: 9,  day: 5  },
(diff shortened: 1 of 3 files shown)


test/helpers/holiday_helpers.rb
```diff
@@ -0,0 +1,46 @@
+module HolidayHelpers
+  class MissingConfigurationError < RuntimeError; end
+
+  HOLIDAYS = {
+    ups: { # https://compass.ups.com/ups-holiday-schedule-2016/
+      "2016" => [
+        [1,  1 ],
```
MalazAlamir: suggestion: Change this to a hash with `month` and `day` keys? 

I don't feel strongly about this suggestion but could be more readable.

Also, having such a complex data structure makes me feel like there should be a `Holidays` object or something, but maybe thats an overkill for these tests.

benwah:
squashed..

test/helpers/holiday_helpers.rb
benwah: I went along with your suggestion @MalazAlamir

PR #453, opened 2017-01-01
Deprecate Fixnum usage for Integer (Ruby 2.4 Compatibility)

ddrmanxbxfr:
In ruby 2.4, Fixnum and Bignums were unified together into the Integer class.

We should deprecate them so that we can upgrade to ruby 2.4 without deprecations.

Links related to this change in ruby...

https://bugs.ruby-lang.org/issues/12005
http://blog.bigbinary.com/2016/11/18/ruby-2-4-unifies-fixnum-and-bignum-into-integer.html

### DIFF
--- a/lib/active_shipping/package.rb
+++ b/lib/active_shipping/package.rb
@@ -142,7 +142,7 @@ module ActiveShipping #:nodoc:
 
     def measure(measurement, ary)
       case measurement
-      when Fixnum then ary[measurement]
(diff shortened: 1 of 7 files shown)


ddrmanxbxfr:
@kmcphillips done, I've rebased against master.

thegedge:
Thanks, @ddrmanxbxfr!

PR #456, opened 2017-01-04
Add Ruby 2.4.0 to CI

kmcphillips:
Ruby 2.4 has some behaviour changes, most notable the `Integer` class combining `Fixnum` and `Bignum`:
https://www.ruby-lang.org/en/news/2016/12/25/ruby-2-4-0-released/

This adds it to CI. 

Related to #453. cc @ddrmanxbxfr

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,11 +5,17 @@ sudo: false
 rvm:
 - "2.2.2"
 - "2.3.1"
+- "2.4.0"
 
 gemfile:
 - gemfiles/activesupport42.gemfile
 - gemfiles/activesupport50.gemfile
 - gemfiles/activesupport_master.gemfile
+
+matrix:
+  exclude:
+    - rvm: "2.4.0"
+      gemfile: gemfiles/activesupport42.gemfile
 
 env:
   global:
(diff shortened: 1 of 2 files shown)


test/unit/carriers/new_zealand_post_test.rb
```diff
@@ -145,7 +145,6 @@ def test_international_empty_json_response_error
     @carrier.expects(:commit).returns([""])
     error = @carrier.find_rates(@wellington, @ottawa, package_fixtures[:book]) rescue $!
     assert_equal ActiveShipping::ResponseError, error.class
-    assert_equal "A JSON text must at least contain two octets!", error.message
```
kmcphillips: This error message is tied to the version of JSON being used. And fails across the 1.x 2.x version change.

.travis.yml
```diff
@@ -5,12 +5,18 @@ sudo: false
...
 
 gemfile:
 - gemfiles/activesupport42.gemfile
 - gemfiles/activesupport50.gemfile
 - gemfiles/activesupport_master.gemfile
 
+matrix:
+  exclude:
+    - rvm: "2.4.0"
+      gemfile: gemfiles/activesupport42.gemfile
```
kmcphillips: Cannot install JSON which is used by 4.2 at this version. Skip this permutation.

Issue #455, opened 2017-01-03
Bug: Stamps.com rates are expressed in dollars in the API, ActiveShipping sometimes interprets them as cents


dustMason:
Before I make a PR, I'd like advice on the best approach to fixing this issue.

The code in question is https://github.com/Shopify/active_shipping/blob/master/lib/active_shipping/package.rb#L117-L131

When querying the Stamps.com API for rates, it returns amount values that look like `10.50` representing $10.50, for example. However, when the value is an amount of whole dollars, it looks like `10` representing $10, for example. This causes the `cents_from` method above to assume that the `10` is a value in cents since it doesn't contain a `.`.

Perhaps the best way would be to have the `Stamps` Carrier class return this value as an instance of a class that implements a `cents` method. What would be best practice here?

Thanks

kmcphillips:
Oh boy. This `cents_from` method is super painful.

I'm going to put this in the 2.0 milestone. We should replace this with proper money parsing.

PR #457, opened 2017-01-05
Quick fix for Stamps.com rate amount bug

dustMason:
In lieu of rewriting `ActiveShipping::Package#cents_from` as discussed in #455, this is an effective fix for the bug described on that issue.


dustMason:
I returned to this just now to add a test and realized that there are money amounts in several places in the Stamps.com API responses that should be treated the same (as decimal values).

This, plus the 😟 feeling I get whenever a money value gets converted into a Float, is making me want to re-think this fix. I'm going to close this PR and consider a better fix for the issue.

I think this means that each Carrier subclass should be able to define its own money parsing methods. I'm not sure how else to determine that a given integer value is an amount in dollars or cents.

kmcphillips:
Sound good. Thanks for thinking about quality.

I actually really like the idea of making each Carrier subclass override or implement its own money. cc @jonathankwok @thegedge

PR #463, opened 2017-01-11
Modernize RateEstimateTest

jonathankwok:
1. Use `test "" do` syntax.
2. That's pretty much it. 

I probably could've put this in the same PR as #462

### DIFF
--- a/test/unit/rate_estimate_test.rb
+++ b/test/unit/rate_estimate_test.rb
@@ -1,7 +1,7 @@
 require 'test_helper'
 
 class RateEstimateTest < ActiveSupport::TestCase
-  def setup
+  setup do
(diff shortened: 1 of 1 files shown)


test/unit/rate_estimate_test.rb
```diff
@@ -26,44 +22,40 @@ def test_phone_required
...
   end
 
-  def test_accepts_description_field
+  test "#initialize accepts description option field" do
     rate_estimate = RateEstimate.new(@origin, @destination, @carrier, @service_name, @options.merge(description: "It's free!"))
     assert_equal "It's free!", rate_estimate.description
   end
 
-  def test_date_for_invalid_string_in_ruby_19
-    assert_nil @rate_estimate.send(:date_for, "Up to 2 weeks") if RUBY_VERSION.include?('1.9')
```
kmcphillips: RIP 1.9

benwah: 💀

PR #468, opened 2017-01-12
Adds notion of whether parsed timestamp for ShipmentEvent has a timezone

qq99:
Information around timezones can be lost when creating tracking events.  We've attempted to normalize the shipment event time in tracking events to a "zoneless" fashion, but from an API consumer perspective, it can be handy to know if the UTC time value on the event is really UTC, or UTC only by virtue of the carrier not having supplied a timezone with their timestamp.

Examples:
USPS timestamps look a bit like:
```xml
      <EventTime>9:01 am</EventTime>
      <EventDate>April 28, 2015</EventDate>
```
with no other info as to convey the timezone.  It's probably based on the city/province/country in question, but perhaps it could be USPS server time.  In any case, it's ambiguous and `timezone_ambiguous?` would&should be true.

FedEx timestamps (most? all?) look like this:
```xml
<Timestamp>2014-11-18T18:13:00+11:00</Timestamp>
```
since the TZ is included in the timestamp, `timezone_ambiguous?` would&should be `false`.

This PR adds a method to be more sure whether an instance of `ShipmentEvent` is really zoneless or not

### DIFF
--- a/lib/active_shipping/carriers/fedex.rb
+++ b/lib/active_shipping/carriers/fedex.rb
@@ -662,10 +662,12 @@ module ActiveShipping
           description = event.at('EventDescription').text
           type_code = event.at('EventType').text
 
-          time          = Time.parse(event.at('Timestamp').text)
+          timestamp     = event.at('Timestamp').text
+          time          = Time.parse(timestamp)
+          zoneless      = (timestamp =~ /[-+]\d{2}:\d{2}$/).blank?
           zoneless_time = time.utc
 
-          shipment_events << ShipmentEvent.new(description, zoneless_time, location, description, type_code)
+          shipment_events << ShipmentEvent.new(description, zoneless_time, location, description, type_code, zoneless)
         end
         shipment_events = shipment_events.sort_by(&:time)
 
--- a/lib/active_shipping/carriers/fedex.rb
+++ b/lib/active_shipping/carriers/fedex.rb
@@ -664,12 +664,12 @@ module ActiveShipping
           description = event.at('EventDescription').text
           type_code = event.at('EventType').text
 
-          timestamp     = event.at('Timestamp').text
-          time          = Time.parse(timestamp)
-          zoneless      = (timestamp =~ TIMEZONE_OFFSET_FROM_DATE_REGEX).blank?
-          zoneless_time = time.utc
+          timestamp          = event.at('Timestamp').text
+          time               = Time.parse(timestamp)
+          timezone_ambiguous = (timestamp =~ TIMEZONE_OFFSET_FROM_DATE_REGEX).blank?
+          zoneless_time      = time.utc
 
-          shipment_events << ShipmentEvent.new(description, zoneless_time, location, description, type_code, zoneless)
+          shipment_events << ShipmentEvent.new(description, zoneless_time, location, description, type_code, timezone_ambiguous)
         end
         shipment_events = shipment_events.sort_by(&:time)
 
--- a/lib/active_shipping/carriers/fedex.rb
+++ b/lib/active_shipping/carriers/fedex.rb
@@ -13,6 +13,8 @@ module ActiveShipping
     TEST_URL = 'https://gatewaybeta.fedex.com:443/xml'
     LIVE_URL = 'https://gateway.fedex.com:443/xml'
 
+    TIMEZONE_OFFSET_FROM_DATE_REGEX = /[-+](\d{2}:\d{2})$/
+
     CARRIER_CODES = {
       "fedex_ground" => "FDXG",
       "fedex_express" => "FDXE"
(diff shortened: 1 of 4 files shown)


lib/active_shipping/carriers/fedex.rb
```diff
@@ -662,10 +662,12 @@ def parse_tracking_response(response, options)
           description = event.at('EventDescription').text
           type_code = event.at('EventType').text
 
-          time          = Time.parse(event.at('Timestamp').text)
+          timestamp     = event.at('Timestamp').text
+          time          = Time.parse(timestamp)
+          zoneless      = (timestamp =~ /[-+]\d{2}:\d{2}$/).blank?
```
kmcphillips: I'm a fan of extracting regex into a named constant which explains what it does `TIMEZONE_FROM_DATE_REGEX`.

kmcphillips:
@qq99 We can backport this to 1.x when it's merged.

qq99:
@jonathankwok any thoughts on naming method?
`uses_timezone` and `has_timezone` bug me a bit, since technically it's a `Time` instance so it always has some notion of a timezone on it.  Maybe `timezone_unambiguous?` or `has_known_timezone?`?

qq99:
reping for reviews

lib/active_shipping/carriers/fedex.rb
```diff
@@ -662,10 +664,12 @@ def parse_tracking_response(response, options)
           description = event.at('EventDescription').text
           type_code = event.at('EventType').text
 
-          time          = Time.parse(event.at('Timestamp').text)
-          zoneless_time = time.utc
+          timestamp          = event.at('Timestamp').text
+          time               = Time.parse(timestamp)
+          timezone_ambiguous = (timestamp =~ TIMEZONE_OFFSET_FROM_DATE_REGEX).blank?
```
kmcphillips: 👌 😐

lib/active_shipping/carriers/fedex.rb
```diff
@@ -662,10 +664,12 @@ def parse_tracking_response(response, options)
...
+          zoneless_time      = time.utc
 
-          shipment_events << ShipmentEvent.new(description, zoneless_time, location, description, type_code)
+          shipment_events << ShipmentEvent.new(description, zoneless_time, location, description, type_code, zone_ambiguous)
```
kmcphillips: Pick one or the other. `zone_ambiguous` or (preferably) `timezone_ambiguous`. because here you reference a variable that doesn't exist.

kmcphillips: Throughout all files.

qq99: doh. lol, I thought I had found+replaced this properly, will fix

qq99:
Alright, all should be well now

lib/active_shipping/shipment_event.rb
```diff
@@ -2,14 +2,18 @@ module ActiveShipping
   class ShipmentEvent
     attr_reader :name, :time, :location, :message, :type_code
 
-    def initialize(name, time, location, message = nil, type_code = nil)
-      @name, @time, @location, @message, @type_code = name, time, location, message, type_code
+    def initialize(name, time, location, message = nil, type_code = nil, timezone_ambiguous = true)
```
kmcphillips: Should this still default to true? I think we assume it is _not_ ambiguous by default right? For carriers that do not set this. We don't really test that path.

jonathankwok: [USPS is ambiguous.](https://github.com/Shopify/active_shipping/blob/master/lib/active_shipping/carriers/usps.rb#L263)
[UPS as well.](https://github.com/Shopify/active_shipping/blob/master/lib/active_shipping/carriers/ups.rb#L874)

Worth auditing them all though. I think Canada Post _sometimes_ gives the timezone.

jonathankwok: Looks like Canada Post _does_ use it though. https://github.com/Shopify/active_shipping/blob/master/lib/active_shipping/carriers/canada_post_pws.rb#L325

qq99: I'd say it should be up to the implementer of the parser to mark the timezone as unambiguous.  USPS simply copies across the time integers into a blank `Time.utc(...)`, other trackers we've implemented are also ambiguous.  Many other carriers are building their ShipmentEvents with `zoneless_time` as an argument.

So it feels better to err on the side of being ambiguous over not?

What about `nil` as the default?  Sucks for `ambiguous_zone?` to potentially return 3 values, but I don't feel super confident that the ambiguity is gone from the other carriers untouched in this PR

qq99: [Stamps](https://github.com/Shopify/active_shipping/blob/d842dc36141570f332937dc89b93d5d0e5763b6f/lib/active_shipping/carriers/stamps.rb#L731-L740) ambiguous too

qq99:
Spoke with and handed this off to @jonathankwok since I need to prioritize some other work.

I think we all agree we should properly note timezone ambiguity for each carrier doing tracking events before pushing this.  IMO this should be considered low priority since any new occurrences of the underlying issue are related to DST changeover time (old tracking numbers probably long irrelevant).

We should get it fixed up a bit before the next major DST changeover date, which sounds like it might be around Mar 12, 2017 for most of continental US

qq99:
I think we can close this, @nicolaslupien found an elegant solution to handle these times differently, for any future readers see https://github.com/rails/rails/blob/e0397b0c25e790eba0c914d9ff97360219f49e29/activesupport/lib/active_support/time_with_zone.rb#L472-L478

Issue #466, opened 2017-01-12
Refactor and remove the weight/measurement methods with params on Package and PackageItem


kmcphillips:
Both the `Package` and `PackageItem` classes have methods to return weight and dimensions in all kinds of units. Moreover, those methods accept options hashes and parameters to modify the return values of those methods.

The latter is bad practice. But now with using `measured` we can actually remove _all_ of those methods I believe. Replace them with simply `weight` and other methods for volumetric and dimensional calculations. Then have them return `measured` objects which understand how to convert to different units by themselves.

Should make the files much much much simpler, and easier to work with.

jonathankwok:
Volumetric/Dimensional weight can't be calculated generically - it depends on the carrier's (and possibly their service's) density factor constant.

[USPS has a constant 6000g/cm3](http://www.stamps.com/whitepapers/how-to-calculate-dim-weight-guide.pdf)

Canada Post has it different by service: [6000 for Expedited/Regular Parcel and 5000 for everything else](https://www.canadapost.ca/common/tools/pg/manual/PGabcmail-e.asp)

These functions are generally useless (I know we don't use it at all, instead opting to manually calculate it external to ActiveShipping), and potentially misleading. I'll start by removing them and re-implement them in a carrier-specific context.

Semi-related: [the money parsing per carrier issue](https://github.com/Shopify/active_shipping/issues/458). I think solving this will provide a solution for this as well.

kmcphillips:
I'm really good with pushing these into the carrier implementations, and maybe having a default globally if it makes sense.

garethson:
I'm going to remove this item from the 2.0 milestone, for now.

Issue #458, opened 2017-01-09
Fix money parsing


kmcphillips:
`Package.cents_from` is a problem.
https://github.com/Shopify/active_shipping/blob/e76c8c7314f7fbc6a3f7e66fa076624498c8bf39/lib/active_shipping/package.rb#L117-L131

It makes a bunch of guesses of how it thinks money is being represented, then parses with switch cases and some weird defaults. Plus assumes anything that responds to `cents` is money. 

We could make each `Carrier` implement its own money parser, based on the types coming back for a given provider. Or something cleaner. Plus it shouldn't really be in the `Package` class anyway.

See #457 #455 for examples.

dustMason:
I started working on this the other day. It's a big refactor, made a little harder by the fact that the expected money format of each Carrier is not codified anywhere in this library. We will have to check the docs for each API to make sure we get it right.

Pulling in something like RubyMoney would be overkill, but an `ActiveShipping::Money` class would save us from some money arithmetic pitfalls in the future. Here's what it could look like:

https://gist.github.com/dustMason/433437f0746ba54a07f84f2f8d1dc76a

Then, as mentioned above, each `Carrier` class would create instances of `ActiveShipping::Money` as needed, taking care to parse values in API responses as amount in dollars, cents or whatever.

PR #469, opened 2017-01-12
Change Package interface for ActiveShipping 2.0

jonathankwok:
# Preface
This PR makes major opinionated changes to the interface of `ActiveShipping::Package`. Feel free to chime in with your own opinions, because these opinions are often my own and I've been known to be wrong in the past. 😛 

# Changes
There are many changes here. Here's a list, with some more explicit details below each.

1. The constructor method for `Package` is much more explicit, at the cost of an increase in verbosity. 
    - No more memorization of argument order - `length:, width:, height:, weight:, options:` in any order. Keyword arguments!
2. The only dimension that can be omitted is `Height`.
    - Omitting any dimension raises an `ArgumentError` except for `Height`, which is commonly omitted in the shipping industry (e.g. envelopes!). In this case, it'll default to `Measured::Length(0, :cm)`, where the unit is the same as the defined dimensions' units.
3. `units`/`weight_unit_system`/`dimensions_unit_system`/all the unit things are gone. We're using `Measured` objects so we shouldn't need them.
    - It also means you can mix and match your units willy-nilly. If you want to have a length in centimetres and a width in inches, go wild. 
4. The named unit methods like `ounces`, `pounds`, `centimetres`, `inches`, etc are all gone.
    - I've introduced a `dimensions` method that operates like the old `inches/centimetres` methods, returning an array of the dimensions; only in `Measured` objects, not Floats.
    - `ounces` and `pounds` were mostly just arbitrary shortcuts to converting `Measured` objects and getting their raw values. I don't believe it's necessary but I can be persuaded otherwise. 
5. Volumetric and dimensional weight calculations are removed, because they're calculated on a per-carrier basis, not generically. 
    - See: https://github.com/Shopify/active_shipping/issues/466#issuecomment-272217277

# This is incomplete!
There's a few things that need to be done before this can be merged.

1. Fix everything that breaks because of the interface change. Like 40% of our tests are now red. 
2. Figure out a fix for `cents_from`. Related: #458 
3. Reintroduce volumetric/dimensional weight. Probably the same way we do 2. 
4. Add `Measured 2.0` so we can actually use the new Measured features.

### DIFF
--- a/lib/active_shipping/package.rb
+++ b/lib/active_shipping/package.rb
@@ -1,118 +1,62 @@
 module ActiveShipping #:nodoc:
   class Package
-    cattr_accessor :default_options
-    attr_reader :options, :value, :currency
+    cattr_accessor :default_options do {} end
+    attr_reader :weight, :length, :width, :height, :options, :value, :currency
 
-    # Package.new(100, [10, 20, 30], :units => :metric)
-    # Package.new(Mass.new(100, :grams), [10, 20, 30].map {|m| Length.new(m, :centimetres)})
-    # Package.new(100.grams, [10, 20, 30].map(&:centimetres))
-    def initialize(grams_or_ounces, dimensions, options = {})
-      options = @@default_options.update(options) if @@default_options
-      options.symbolize_keys!
-      @options = options
+    alias_attribute :mass, :weight
 
-      @dimensions = [dimensions].flatten.reject(&:nil?)
+    def initialize(weight:, length:, width:, height: nil, options: {})
+      @options = @@default_options.merge(options.symbolize_keys)
 
-      imperial = (options[:units] == :imperial) ||
-                 ([grams_or_ounces, *dimensions].all? { |m| m.respond_to?(:unit) && m.unit.to_sym == :imperial })
+      raise ArgumentError, "Weight needs to be a Measured::Measurable object" unless weight.is_a?(Measured::Measurable)
+      raise ArgumentError, "Length needs to be a Measured::Measurable object" unless length.is_a?(Measured::Measurable)
+      raise ArgumentError, "Width needs to be a Measured::Measurable object" unless width.is_a?(Measured::Measurable)
+      raise ArgumentError, "Height needs to be a Measured::Measurable object" unless height.nil? || length.is_a?(Measured::Measurable)
 
-      weight_imperial = dimensions_imperial = imperial if options.include?(:units)
+      @weight = weight
 
-      if options.include?(:weight_units)
-        weight_imperial = (options[:weight_units] == :imperial) ||
-                          (grams_or_ounces.respond_to?(:unit) && m.unit.to_sym == :imperial)
-      end
+      @length = length
+      @width  = width
+      @height = height || ( cylinder? ? width : Measured::Length.new(0, length.unit) )
 
-      if options.include?(:dim_units)
-        dimensions_imperial = (options[:dim_units] == :imperial) ||
-                              (dimensions && dimensions.all? { |m| m.respond_to?(:unit) && m.unit.to_sym == :imperial })
-      end
+      @value = self.class.cents_from(options[:value])
+      @currency = options[:currency] || options[:value].try(:currency)
+    end
 
-      @weight_unit_system = weight_imperial ? :imperial : :metric
-      @dimensions_unit_system = dimensions_imperial ? :imperial : :metric
+    def dimensions
+      [@length, @width, @height]
+    end
 
-      @weight = attribute_from_metric_or_imperial(grams_or_ounces, Measured::Weight, @weight_unit_system, :grams, :ounces)
+    def girth
+      @girth ||= cylinder ? width.scale(Math::PI) : (width + height).scale(2)
+    end
+    alias_method :circumference, :girth
+    alias_method :around, :girth
 
-      if @dimensions.blank?
-        zero_length = Measured::Length.new(0, (dimensions_imperial ? :inches : :centimetres))
-        @dimensions = [zero_length] * 3
+    def volume
+      @volume ||= if cylinder?
+        Math::PI * width.scale(0.5).value * height.scale(0.5).value * length.value
       else
-        process_dimensions
+        length.value * width.value * height.value
       end
-
-      @value = Package.cents_from(options[:value])
-      @currency = options[:currency] || (options[:value].currency if options[:value].respond_to?(:currency))
-      @cylinder = (options[:cylinder] || options[:tube]) ? true : false
-      @gift = options[:gift] ? true : false
-      @oversized = options[:oversized] ? true : false
-      @unpackaged = options[:unpackaged] ? true : false
-    end
-
-    def unpackaged?
-      @unpackaged
-    end
-
-    def oversized?
-      @oversized
     end
 
     def cylinder?
-      @cylinder
+      @cylinder ||= @options[:cylinder].present?
     end
     alias_method :tube?, :cylinder?
 
     def gift?
-      @gift
+      @gift ||= @options[:gift].present?
     end
 
-    def ounces(options = {})
-      weight(options).convert_to(:oz).value
+    def oversized?
+      @oversized ||= @options[:oversized].present?
     end
-    alias_method :oz, :ounces
 
-    def grams(options = {})
-      weight(options).convert_to(:g).value
+    def unpackaged?
+      @unpackaged ||= @options[:unpackaged].present?
     end
-    alias_method :g, :grams
-
-    def pounds(options = {})
-      weight(options).convert_to(:lb).value
-    end
-    alias_method :lb, :pounds
-    alias_method :lbs, :pounds
-
-    def kilograms(options = {})
-      weight(options).convert_to(:kg).value
-    end
-    alias_method :kg, :kilograms
-    alias_method :kgs, :kilograms
-
-    def inches(measurement = nil)
-      @inches ||= @dimensions.map { |m| m.convert_to(:in).value }
-      measurement.nil? ? @inches : measure(measurement, @inches)
-    end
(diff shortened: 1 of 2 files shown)


lib/active_shipping/package.rb
```diff
@@ -1,118 +1,62 @@
 module ActiveShipping #:nodoc:
   class Package
-    cattr_accessor :default_options
-    attr_reader :options, :value, :currency
+    cattr_accessor :default_options do {} end
```
kmcphillips: Prefer:

```ruby
cattr_accessor: default_options
self.default_options = {}
```

lib/active_shipping/package.rb
```diff
@@ -1,118 +1,62 @@
...
+    end
 
-      @value = Package.cents_from(options[:value])
-      @currency = options[:currency] || (options[:value].currency if options[:value].respond_to?(:currency))
-      @cylinder = (options[:cylinder] || options[:tube]) ? true : false
-      @gift = options[:gift] ? true : false
-      @oversized = options[:oversized] ? true : false
-      @unpackaged = options[:unpackaged] ? true : false
+    def dimensions
+      [@length, @width, @height]
```
kmcphillips: This can be memoized too, if this object is correctly immutable.

lib/active_shipping/package.rb
```diff
@@ -1,118 +1,62 @@
...
+      @girth ||= cylinder ? width.scale(Math::PI) : (width + height).scale(2)
     end
+    alias_method :circumference, :girth
+    alias_method :around, :girth
 
-    def oversized?
-      @oversized
+    def volume
+      @volume ||= if cylinder?
+        Math::PI * width.scale(0.5).value * height.scale(0.5).value * length.value
```
kmcphillips: 👌 dat-`scale`

kmcphillips: But, you're switching to `value` here. And that makes the incorrect assumption that `unit` is the same on all of these.

lib/active_shipping/package.rb
```diff
@@ -1,118 +1,62 @@
...
+      else
+        length.value * width.value * height.value
+      end
     end
 
     def cylinder?
-      @cylinder
+      @cylinder ||= @options[:cylinder].present?
```
kmcphillips: This is then saying `cylinder: "false"` means that it _is_ a cylinder.

kmcphillips: Not necessarily wrong, but as long as we're ok with that. And document it. And are consistent.

test/unit/package_test.rb
```diff
@@ -2,403 +2,145 @@
 
 class PackageTest < ActiveSupport::TestCase
   setup do
-    @weight = 100
-    @dimensions = [5, 6, 7]
+    @weight = Measured::Weight(500, :grams)
+    @length = Measured::Length(30, :cm)
+    @width = Measured::Length(25, :cm)
+    @height = Measured::Length(15, :cm)
```
kmcphillips: Switch one of these to another unit. That would have failed the `.value` bug I pointed out above.

test/unit/package_test.rb
```diff
@@ -2,403 +2,145 @@
...
-  test "#inches alias to #in" do
-    assert_equal @dimensions, @imperial_package.inches
-    assert_equal @dimensions, @imperial_package.in
+  test "#initialize with no options defaults them all to false" do
+    package = Package.new(**@default_args.merge(options: {}))
+    refute_predicate package, :cylinder?
+    refute_predicate package, :tube?
+    refute_predicate package, :oversized?
+    refute_predicate package, :unpackaged?
+    refute_predicate package, :gift?
```
kmcphillips: Thanks for changing to `_predicate`

test/unit/package_test.rb
```diff
@@ -2,403 +2,145 @@
...
+  test "#volume returns the box volume for non-cylindrical package" do
+    box_volume = @length.value * @width.value * @height.value
+    assert_equal box_volume, @default_package.volume
   end
 
-  private
+  # This should return a Measured::Volume, but that doesn't exist yet.
+  # So we'll return a raw value like we have in the past until that is a thing.
+  test "#volume returns the volume for cylindrical package" do
+    skip "Requires Measured 2.0"
```
kmcphillips:
```ruby
if Gem::Version.new(Measured::VERSION) < Gem::Version.new("2.0.0.pre1")
  skip "Requires Measured 2.0"
end
```

lib/active_shipping/package.rb
thegedge: Two things:

First, would the measured gem value from a function to coerce a list of values into the same unit? For example, this would be written as:

```ruby
Measured.coerce_unit(width.scale(0.5), height.scale(0.5), length).reduce(&:*)
```

We could also add a function to do the inner product of the values until we support multiplying these together.

Second, I would recommend parentheses around the three terms on the right, because the first multiply will coerce things into a float and then you'll lose some precision gained from `:value` being `Rational` or `BigDecimal`. The other option would be to create a `Rational` or `BigDecimal` PI constant in this class and use that instead (recommended). The nice thing about that is that it will reduce the number of return value types.

lib/active_shipping/package.rb
```diff
@@ -1,118 +1,62 @@
...
-    # Package.new(100.grams, [10, 20, 30].map(&:centimetres))
-    def initialize(grams_or_ounces, dimensions, options = {})
-      options = @@default_options.update(options) if @@default_options
-      options.symbolize_keys!
-      @options = options
+    alias_attribute :mass, :weight
 
-      @dimensions = [dimensions].flatten.reject(&:nil?)
+    def initialize(weight:, length:, width:, height: nil, options: {})
+      @options = @@default_options.merge(options.symbolize_keys)
```
thegedge: If `@@default_options` can be changed by a client of this library then I'd recommend `@@default_options.merge(options).symbolize_keys` to guarantee all keys are symbols.

lib/active_shipping/package.rb
```diff
@@ -1,118 +1,62 @@
...
+      @height = height || ( cylinder? ? width : Measured::Length.new(0, length.unit) )
 
-      if @dimensions.blank?
-        zero_length = Measured::Length.new(0, (dimensions_imperial ? :inches : :centimetres))
-        @dimensions = [zero_length] * 3
-      else
-        process_dimensions
-      end
+      @value = self.class.cents_from(options[:value])
+      @currency = options[:currency] || options[:value].try(:currency)
```
thegedge: What if I pass in the following `Hash` for options:
```ruby
{
  value: MoneyWithCurrency.new(10, currency: :usd),
  currency: :cad
}
```

lib/active_shipping/package.rb
```diff
@@ -1,118 +1,62 @@
...
-      @inches ||= @dimensions.map { |m| m.convert_to(:in).value }
-      measurement.nil? ? @inches : measure(measurement, @inches)
-    end
-    alias_method :in, :inches
-
-    def centimetres(measurement = nil)
-      @centimetres ||= @dimensions.map { |m| m.convert_to(:cm).value }
-      measurement.nil? ? @centimetres : measure(measurement, @centimetres)
+    def oversized?
+      @oversized ||= @options[:oversized].present?
```
mdking: Again I think we should look at what this is describing. It's a carrier's view of physical rating characteristics of a package.

e.g. for Canada Post it might be
```
options = {
  rating_characteristics: [:oversized]
}
```

Where as for USPS it might be

```
options = {
  rating_characteristics: [:oversized, :cubic]
```

Thoughts?

lib/active_shipping/package.rb
jonathankwok: Great insight. I'm okay with `shape`, a Package generally has a shape - box, tube, or envelope. I won't enforce the enumeration of the shape, but I'll keep the boolean methods for common shapes e.g. box/tube/envelope.

lib/active_shipping/package.rb
```diff
@@ -1,118 +1,62 @@
...
 
     def cylinder?
-      @cylinder
+      @cylinder ||= @options[:cylinder].present?
     end
     alias_method :tube?, :cylinder?
 
     def gift?
-      @gift
+      @gift ||= @options[:gift].present?
```
jonathankwok: ~~`export_reason` is a little strangely worded to me, mostly for the `export` included. Maybe `mail_reason`? `ship_reason`?~~

Actually, since this is a customs concern, `export` is perfectly natural. I'll go with `export_reason` unless somebody has arguments for otherwise.

lib/active_shipping/package.rb
jonathankwok: That's a bit wordy, but best describes it since `characteristics` is too vague.

I'm okay with `rating_classifications` as well.

lib/active_shipping/package.rb
thegedge: I think I would actually trust the currency attached to `value` more than a given currency, but in the example I gave there's an inconsistency, so I think it would be fair to raise `ArgumentError`.

jonathankwok:
Had a great conversation with @mdking about rating concerns vs generic package concerns. TLDR

 - `rating_classification` shouldn't be an attribute on a Package, that's a _rating_ concern. It should be on a RateEstimate
 - `export_reason` is the same, for the reasons above. A Package independently has no export reason. It should be attached to either a RateEstimate or the Shipment that holds the package.
 - `shape` can be kept on a package. It still describes a Package independent of the rate/shipment. 

Will make these changes and iterate, but as always, everything is fluid and additional feedback is super welcome.

kevin-jj:
What about this PR in the version 2.0? @jonathankwok @anthonyn60

jonathankwok:
I don't quite have the time to work on this, so if somebody wants to see this merged they can pick up where I left off somewhere in a new PR.

PR #473, opened 2017-01-20
Add encrypted UPS credentials

jonathankwok:
I got some credentials to run remote tests for UPS. Using the [encrypted keys](https://docs.travis-ci.com/user/encryption-keys/) feature in Travis, hopefully this should let us run remote tests without publicly exposing sensitive credentials.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,22 +1,18 @@
 language: ruby
 script: bundle exec rake test
 sudo: false
-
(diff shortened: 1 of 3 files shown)


jonathankwok:
Okay this doesn't work, you're correct - `puts CREDENTIALS` will reveal the credentials. So close!

PR #476, opened 2017-01-23
Raise a ResponseError for FedEx#find_tracking_info

thegedge:
In a couple of places we were raising an `ActiveShipping::Error` when we could raise/return a more informed `ActiveShipping::TrackingResponse(false, ...)` that has additional information (e..g, response, request) for clients of this gem.

The one point of discussion here is w.r.t. the semantics of `ResponseError`. Generally it means there was an issue in the response, but for these two cases I've changed aren't considered error responses by FedEx, and aren't given special status codes (i.e., `/TrackReply/Notifications/Code` is 0), so perhaps this doesn't make sense?

Also made a few small formatting changes on nearby code (separate commit).

### DIFF
--- a/test/unit/carriers/fedex_test.rb
+++ b/test/unit/carriers/fedex_test.rb
@@ -318,14 +318,31 @@ class FedExTest < ActiveSupport::TestCase
     @carrier.expects(:commit).returns(mock_response)
 
     assert_raises ActiveShipping::ResponseContentError do
-    @carrier.find_rates(
+      @carrier.find_rates(
         location_fixtures[:ottawa],
         location_fixtures[:beverly_hills],
         package_fixtures.values_at(:book, :wii),
-        :test => true
+        test: true
       )
     end
   end
+
+  def test_parsing_response_without_notifications
+    mock_response = xml_fixture('fedex/reply_without_notifications')
+
+    @carrier.expects(:commit).returns(mock_response)
+
+    response = @carrier.find_rates(
+      location_fixtures[:ottawa],
+      location_fixtures[:beverly_hills],
+      package_fixtures.values_at(:book, :wii),
+      test: true
+    )
+
+    assert_predicate response, :success?
+  end
+
+  ### find_tracking_info
 
   def test_response_transient_failure
     mock_response = xml_fixture('fedex/tracking_response_failure_code_9045')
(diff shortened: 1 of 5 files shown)


test/unit/carriers/fedex_test.rb
```diff
@@ -327,6 +327,23 @@ def test_failure_to_parse_invalid_xml_results_in_a_useful_error
...
+  def test_parsing_response_without_notifications
+    mock_response = xml_fixture('fedex/reply_without_notifications')
+
+    @carrier.expects(:commit).returns(mock_response)
+
+    response = @carrier.find_rates(
+      location_fixtures[:ottawa],
+      location_fixtures[:beverly_hills],
+      package_fixtures.values_at(:book, :wii),
+      :test => true
```
jonathankwok: `test: true` would be 👍  here.

test/unit/carriers/fedex_test.rb
```diff
@@ -327,6 +327,23 @@ def test_failure_to_parse_invalid_xml_results_in_a_useful_error
...
+    )
+
+    assert response.success?
```
jonathankwok: `assert_predicate`

test/unit/carriers/fedex_test.rb
```diff
@@ -507,14 +507,25 @@ def test_tracking_info_without_status
...
-    error = assert_raises(ActiveShipping::Error) do
+    error = assert_raises(ActiveShipping::ResponseError) do
       @carrier.find_tracking_info('123456789012')
     end
 
     msg = 'Multiple matches were found. Specify a unqiue identifier: 2456987000~123456789012~FX, 2456979001~123456789012~FX, 2456979000~123456789012~FX'
     assert_equal msg, error.message
   end
 
+  def test_tracking_info_without_tracking_details
```
jonathankwok:
```
test "#find_tracking_info without tracking details" do
```

jonathankwok: You could also change it throughout this PR. Or, keep it as is and have a dedicated PR to updating this test file to `test "" do`

thegedge: Yeah, I'd rather keep that as a bigger cleanup PR to keep things cohesive.

thegedge:
Updated

thegedge:
Okay, I added one more commit which was the actual real issue that needed to be dealt with here, but we can keep the `TrackingResponse` code here too.

tl;dr we were only looking at the success status from the document root, and in a `TrackDetails` element, but forgetting to check `CompleteTrackDetails`. I've updated the test to capture this case.

thegedge:
^ Squashed it into two, cohesive commits that I'll merge (they don't make sense to merge together)

Issue #472, opened 2017-01-18
UPS > get_delivery_date_estimates > need one more tag called TotalPackagesInShipment


FantasticJimmy:
Hi,

As I was playing with the UPS testing API, specifically the ups.get_delivery_date_estimates.
Error was returned to me saying "weight is invalid." I later discover that this was due to the fact, if no TotalPackagesInShipment tag was given, it is by default assuming that the shipment only contains one package. 
So in the ups.rb file, under function build_delivery_dates_request:

I added for the request to be valid.

If you think this is a valid issue, please modify and update the gem please.

            xml.Weight([value.round(3), 0.1].max)
          end
#################
          **xml.TotalPackagesInShipment(packages.count)**
####################
          if packages.any? {|package| package.value.present?}

Thank you.

Jimmy

jonathankwok:
Hey @FantasticJimmy. Thanks for the insight - up until recently we hadn't handled cases where shipments could have multiple packages, hence the missing `xml.TotalPackagesInShipment`. 

Your change is pretty straightforward and does the job, so if you would like to raise a PR, feel free to do so.

AdnanTheExcellent:
Thank you for bringing this up. I have been trying to fix this problem for a while now and couldnt figure it out until i saw this post.

PR #477, opened 2017-01-23
added TotalPackagesInShipment tag in ups.rb for time_in_transit request

FantasticJimmy:
refer to issue #472 
Resolved.


test/remote/ups_test.rb
```diff
@@ -393,6 +393,22 @@ def test_delivery_date_estimates_intl
...
+    response = @carrier.get_delivery_date_estimates(
+      location_fixtures[:new_york_with_name],
+      location_fixtures[:ottawa_with_name],
+      package_fixtures.values_at(:multiple_packages),
+      pickup_date=today,
+      {
+        :test => true
+      }
+    )
+    #this test is soly to test if the dates_request is valid, test would fail if missing or invalid request parameters.
```
jonathankwok: Comments explaining the purpose of a test are an indicator that the test purpose is unclear. If your test name is eloquent and descriptive, there should be no need for a comment.

test/remote/ups_test.rb
```diff
@@ -393,6 +393,22 @@ def test_delivery_date_estimates_intl
...
+  def test_delivery_date_estimates_to_have_valid_request_when_package_is_heavy_and_multiple_box
+    today = Date.current
+
+    response = @carrier.get_delivery_date_estimates(
+      location_fixtures[:new_york_with_name],
+      location_fixtures[:ottawa_with_name],
+      package_fixtures.values_at(:multiple_packages),
+      pickup_date=today,
+      {
+        :test => true
```
jonathankwok: This doesn't need to be fixed in this PR, since this file is littered with this problem, but for future reference - you can use the hash colon syntax. In addition, the curly braces are unnecessary here, since a terminating hash argument doesn't require braces.

developingchris:
How can I help get this finished up?

PR #478, opened 2017-01-25
Update ups rb file with one tag added called total packages in shipment

FantasticJimmy:
Written a remote UPS test case to check of the time_in_transit request was valid. Since the test was using multiple packages, test would fail if < TotalPackagesInShipment > {number}</TotalPackagesInShipment> tag is missing from the request.


jonathankwok:
This looks like a duplicate of https://github.com/Shopify/active_shipping/pull/477, closing.

PR #481, opened 2017-02-07
Add available_services method to carrier class

nicolaslupien:
# Why

Carriers implement different ways to define what services are available for a route. We should have a common interface to fetch this information.

# What

By adding `available_services` method to the `Carrier` class, we could easily fetch a list of available services between two countries.

Other things to consider: There's a few other things that could impact the service availability like package price and weight (and many more). I figured the most common would be by route (countries from and to) and if you have a package, you can always just fetch the rates to see the available services.

I'm also not sure about the input parameters, using countries seems simple enough and to integrate well with carriers but we could also use a full `Location` to represent the origin and destination.

cc @qq99 @mdking

### DIFF
--- a/lib/active_shipping/carrier.rb
+++ b/lib/active_shipping/carrier.rb
@@ -94,11 +94,11 @@ module ActiveShipping
 
     # Get a list of services available for the a specific route
     #
-    # @param origin_country [String] The country of origin
-    # @param destination_country [String] The destination country
+    # @param origin_country_code [String] The country of origin
+    # @param destination_country_code [String] The destination country
     # @return [Array<String>] A list of names of the available services
     #
-    def available_services(origin_country, destination_country)
+    def available_services(origin_country_code, destination_country_code, options = {})
       raise NotImplementedError, "#available_services is not supported by #{self.class.name}."
     end
 
--- a/lib/active_shipping/carrier.rb
+++ b/lib/active_shipping/carrier.rb
@@ -92,6 +92,16 @@ module ActiveShipping
       raise NotImplementedError, "#find_tracking_info is not supported by #{self.class.name}."
     end
 
+    # Get a list of services available for the a specific route
+    #
+    # @param origin_country [String] The country of origin
+    # @param destination_country [String] The destination country
+    # @return [Array<String>] A list of names of the available services
+    #
+    def available_services(origin_country, destination_country)
+      raise NotImplementedError, "#available_services is not supported by #{self.class.name}."
+    end
+
     # Validate credentials with a call to the API.
     #
     # By default this just does a `find_rates` call with the origin and destination both as
--- a/test/unit/carrier_test.rb
+++ b/test/unit/carrier_test.rb
@@ -34,6 +34,12 @@ class CarrierTest < ActiveSupport::TestCase
     end
   end
 
+  test "#available_services is not implemented" do
+    assert_raises NotImplementedError do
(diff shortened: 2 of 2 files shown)


lib/active_shipping/carrier.rb
```diff
@@ -92,6 +92,16 @@ def find_tracking_info(tracking_number, options = {})
       raise NotImplementedError, "#find_tracking_info is not supported by #{self.class.name}."
     end
 
+    # Get a list of services available for the a specific route
+    #
+    # @param origin_country [String] The country of origin
```
mdking: 1. Can we be explicit in the argument this should be the country code as opposed to the country name. 

e.g. `@param origin_country_code ...`

2. Furthermore call it out as the ISO-alpha 2 which is available from `ActiveUtils`

e.g. `[String] The ISO alpha 2 origin country code`

IMO no. 1 would at least add clarity if they should pass `CA` or `Canada` and not have to deal with doing a look up to convert one to the other.

lib/active_shipping/carrier.rb
```diff
@@ -92,6 +92,16 @@ def find_tracking_info(tracking_number, options = {})
...
+    # @param destination_country [String] The destination country
+    # @return [Array<String>] A list of names of the available services
+    #
+    def available_services(origin_country, destination_country)
```
mdking: Discussion: Add `options = {}` to be consistent and for possible future options such as optionally providing?

lib/active_shipping/carrier.rb
nicolaslupien: I was voluntary being vague here because to follow the steps of `ActiveShipping::Location.country` which can be _either_ a full name or a country code. Should we remove this flexibility here?

mdking: For v2 I think we should have a clear stance on `_name` vs `_code` representations. It's causing headache in our other codes bases. 

@jonathankwok is heading up the vision for v2 and I'd probably say we'd be ok with porting this to the v1.* branch. Kwok, what's your opinion on being explicit about argument representations?

For v2 IMO we should also remove the ambiguity on `ActiveShipping::Location.country`

jonathankwok: Agreed - I want to be clear with using `name` vs `code` suffixes. There's a time and place for a `name`, and for `code`, and in general I'd say the former is for presentation and the latter is for logic.

Reason being, If you follow `ISO 3166-1` standards (or something similar), you'll never run into instances where you have two representations of the same country (e.g. `North Korea` vs `Democratic People's Republic of Korea`, where the code is always `KP`). It doesn't matter whether you call it North Korea or DPRK in your logic, because your logic is only concerned with `KP`. It's only when you surface it that it "matters".

TLDR use `_code`, we'll remove any ambiguity in `Location.country` as well - a general theme in v2.

lib/active_shipping/carrier.rb
jonathankwok: Yeah, as much as I dislike options hashes, they feel like a necessary evil when integrating with carriers in general.

lib/active_shipping/carrier.rb
nicolaslupien: Thanks for the feedback!

nicolaslupien:
@mdking @jonathankwok re-ping for review

jonathankwok:
Backporting this to `v1.x` would be a good idea.

PR #482, opened 2017-02-14
Clean up ShipmentPacker#pack

MalazAlamir:
No functionality change here, just cleaning this method up in an attempt to make it more readable. I've noticed there might be a bug in how we pack some items here, and I will follow up with a fix + explanation. But for now, just prepping to make the change easy :)

### DIFF
--- a/lib/active_shipping/shipment_packer.rb
+++ b/lib/active_shipping/shipment_packer.rb
@@ -11,63 +11,90 @@ module ActiveShipping
     # dimensions      - `[5.0, 15.0, 30.0]`
     # maximum_weight  - maximum weight in grams
     # currency        - ISO currency code
-    def self.pack(items, dimensions, maximum_weight, currency)
-      return [] if items.empty?
-      packages = []
 
-      # Naive in that it assumes weight is equally distributed across all items
-      # Should raise early enough in most cases
-      total_weight = 0
-      items.map!(&:symbolize_keys).each do |item|
-        total_weight += item[:quantity].to_i * item[:grams].to_i
+    class << self
+      def pack(items, dimensions, maximum_weight, currency)
+        return [] if items.empty?
+        packages = []
+        items.map!(&:symbolize_keys)
 
-        if item[:grams].to_i > maximum_weight
-          raise OverweightItem, "The item with weight of #{item[:grams]}g is heavier than the allowable package weight of #{maximum_weight}g"
-        end
+        # Naive in that it assumes weight is equally distributed across all items
+        # Should raise early enough in most cases
+        total_weight = calculate_total_weight(items, maximum_weight)
+        items_to_pack = items.map(&:dup).sort_by! { |i| i[:grams].to_i }
 
-        if total_weight > maximum_weight * EXCESS_PACKAGE_QUANTITY_THRESHOLD
-          raise ExcessPackageQuantity, "Unable to pack more than #{EXCESS_PACKAGE_QUANTITY_THRESHOLD} packages"
-        end
-      end
-
-      items = items.map(&:dup).sort_by! { |i| i[:grams].to_i }
-
-      state = :package_empty
-      while state != :packing_finished
-        case state
-        when :package_empty
-          package_weight, package_value = 0, 0
-          state = :filling_package
-        when :filling_package
-          items.each do |item|
-            quantity = if item[:grams].to_i <= 0
-              item[:quantity].to_i
-            else
-              # Grab the max amount of this item we can fit into this package
-              # Or, if there are fewer than the max for this item, put
-              # what is left into this package
-              [(maximum_weight - package_weight) / item[:grams].to_i, item[:quantity].to_i].min
+        state = :package_empty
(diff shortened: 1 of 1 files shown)


lib/active_shipping/shipment_packer.rb
```diff
@@ -11,63 +11,90 @@ class ExcessPackageQuantity < StandardError; end
...
-              # Or, if there are fewer than the max for this item, put
-              # what is left into this package
-              [(maximum_weight - package_weight) / item[:grams].to_i, item[:quantity].to_i].min
-            end
+      private
 
-            item_weight = quantity * item[:grams].to_i
-            item_value = quantity * Package.cents_from(item[:price])
+      def calculate_total_weight(items, maximum_weight)
+        items.inject(0) do |total_weight, item|
```
jonathankwok: [`ActiveSupport#sum`](http://apidock.com/rails/Enumerable/sum) does the same, and is arguably clearer.

```
items.sum do |item|
  item[:quantity].to_i * items[:grams].to_i
end
```

lib/active_shipping/shipment_packer.rb
```diff
@@ -11,63 +11,90 @@ class ExcessPackageQuantity < StandardError; end
...
+          total_weight
+        end
+      end
+
+      def overweight_item?(grams, maximum_weight)
+        grams.to_i > maximum_weight
+      end
+
+      def excess_package_quantity?(total_weight, maximum_weight)
+        total_weight > (maximum_weight * EXCESS_PACKAGE_QUANTITY_THRESHOLD)
```
jonathankwok: Throwing a question out there: does this make sense?

The error is saying "Unable to pack more than 10000 packages" but the validation is checking if `total_weight > maximum_weight * 10000`. Which one are we validating; # of packages or total weight?

lib/active_shipping/shipment_packer.rb
MalazAlamir: Don't think we can use sum here since we use the `total_weight` in each iteration to do some checks and potentially `raise`

jonathankwok: Never mind, you can't - `sum` doesn't have an accumulator variable to validate against.

lib/active_shipping/shipment_packer.rb
```diff
@@ -11,63 +11,90 @@ class ExcessPackageQuantity < StandardError; end
...
+              quantity = determine_fillable_quantity_for_package(item, maximum_weight, package_weight)
+              package_weight += item_weight(quantity, item[:grams])
+              package_value += item_value(quantity, item[:price])
+              item[:quantity] = item[:quantity].to_i - quantity
+            end
+
+            items.reject! { |i| i[:quantity].to_i == 0 }
+            state = :package_full
+          when :package_full
+            packages << ActiveShipping::Package.new(package_weight, dimensions, :value => package_value, :currency => currency)
```
jonathankwok: Might as well use colon hash syntax.

lib/active_shipping/shipment_packer.rb
nicolaslupien: you can detect those before doing the sum/inject

```ruby
raise if items.any? { |item| overweight_item(item, max)}
total_weight = items.sum? { item[:quantity].to_i * items[:grams].to_i }
raise if excess_package_quantity?(total_weight, maximum_weight)
```

MalazAlamir: I think it's more effecient to do the `raise` for excess quantity in the loop, as it does the check against the accumulator in each iteration and `raise`s earlier than if we were to compute the total weight then `raise`

nicolaslupien: do you have a proof?

MalazAlamir: I've run some really basic benchmarks and the results were the following (`original` is the implementation I had, `suggestion` is with your suggested changes):

With a single item thats overweight:

```
.       user     system      total        real
original  0.000000   0.000000   0.000000 (  0.000008)
suggestion  0.000000   0.000000   0.000000 (  0.000011)
.
```

with large amount of items (1000), each having low weigh 1g, each low quantity of 10:

```
          user     system      total        real
original  0.000000   0.000000   0.000000 (  0.000320)
suggestion  0.000000   0.000000   0.000000 (  0.000359)
..
```

1000 items, low weight, high quantity

```
   .       user     system      total        real
original  0.000000   0.000000   0.000000 (  0.000339)
suggestion  0.000000   0.000000   0.000000 (  0.000359)
.
```

Seems to be almost the same, but original implementation performing slightly better. I'll keep it as is for now.

lib/active_shipping/shipment_packer.rb
MalazAlamir: discussed this on slack but I spoke to blake who originally made this change to this check, and his reasoning was that if the max size of package is 20kg then the most you can ever pack is 20 * 10000 kg, which is 200,000kg, so we'd be packing more than 10,000 packages.

PR #483, opened 2017-02-15
Fix ShipmentPacker#pack to pack whole items when max weight is float

MalazAlamir:
There's currently an issue with `ShipmentPacker#pack` when max weight is passed in as a float. 

https://github.com/Shopify/active_shipping/blob/master/lib/active_shipping/shipment_packer.rb#L49

Particularly `(maximum_weight - package_weight) / item[:grams].to_i`

if max weights is a float (or something like it) they'd be off by one package.. this is because it tries to pack as much as it can in one box, so if they pass a float for `maximum_weight ` we can end up packing half an item

````
irb(main):001:0> 1.0 / 2
=> 0.5
irb(main):002:0> 1 / 2
=> 0
irb(main):003:0> 

````

Which then `items.reject! { |i| i[:quantity].to_i==0 }` discards the remaining half of the item.

Solution is to `.to_i` the numerator, denominator is already an integer.

### DIFF
--- a/lib/active_shipping/shipment_packer.rb
+++ b/lib/active_shipping/shipment_packer.rb
@@ -84,7 +84,8 @@ module ActiveShipping
           # Grab the max amount of this item we can fit into this package
           # Or, if there are fewer than the max for this item, put
           # what is left into this package
-          [(maximum_weight - package_weight) / item_grams, item_quantity].min
+          available_grams = (maximum_weight - package_weight).to_i
+          [available_grams / item_grams, item_quantity].min
         end
       end
 
--- a/test/unit/shipment_packer_test.rb
+++ b/test/unit/shipment_packer_test.rb
@@ -36,6 +36,19 @@ class ShipmentPackerTest < ActiveSupport::TestCase
     end
   end
 
+  def test_divide_order_with_single_line_into_two_packages_max_weight_as_float
+    max_weight = 68038.8555
+
+    items = [{ grams: 45359, quantity: 2, price: 1.0 }]
+
+    packages = ShipmentPacker.pack(items, @dimensions, max_weight, 'USD')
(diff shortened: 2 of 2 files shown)

### CODE REVIEW

File: lib/active_shipping/shipment_packer.rb [hunk 1]
```diff
@@ -84,7 +84,7 @@ def determine_fillable_quantity_for_package(item, maximum_weight, package_weight
           # Grab the max amount of this item we can fit into this package
           # Or, if there are fewer than the max for this item, put
           # what is left into this package
-          [(maximum_weight - package_weight) / item_grams, item_quantity].min
+          [(maximum_weight - package_weight).to_i / item_grams, item_quantity].min
```
jonathankwok: I'd assign `maximum_weight - package_weight` to a variable, so we can have fewer things going on in one line. Also makes it super clear what's happening.

```ruby
available_grams = (maximum_weight - package_weight).to_i
[available_grams / item_grams, item_quantity].min
```

test/unit/shipment_packer_test.rb
```diff
@@ -36,6 +36,19 @@ def test_divide_order_with_single_line_into_two_packages
     end
   end
 
+  def test_divide_order_with_single_line_into_two_packages_max_weight_as_float
+    max_weight = 68038.8555
+
+    items = [{:grams => 45359, :quantity => 2, :price => 1.0}]
```
jonathankwok: I'd advocate for colon syntax! I don't really mind if it's inconsistent within a test file, because when we do come around to amend the rest of the file, it's fewer changes. I could be convinced otherwise though. 😛

MalazAlamir: just thought it looked prettier consistent 😎  but you're right, I'll change it :)

PR #484, opened 2017-02-16
Fix ExcessPackageQuantity validation

MalazAlamir:
The current ExcessPackageQuantity error is checked using:
`total_weight > (maximum_weight * EXCESS_PACKAGE_QUANTITY_THRESHOLD)`

This works for the most part, but this starts to fail when packages are not packed to their max capacity. For example, max weight is 750g, you have one item that weighs 500g with 10001 quantity, the check will do (500 * 10001) > (750 * 10000), which returns false. But at the end of the packing process, you would still end up with 10001 packages, which is over the threshold.

I think this validation is still good, but its not 100% accurate, its a best guess, looser validation. If it returns true, then you will definitely end up with packages over the allowed threshold. But it can miss the case I mentioned above. 

I decided to keep this validation there, to short circuit the the method if `likely_excess_quantity?` returns true (we wont even start the packing process if it returns true), but I also added another validation that simply checks `packages.count` when a package is in full state.

### DIFF
--- a/lib/active_shipping/shipment_packer.rb
+++ b/lib/active_shipping/shipment_packer.rb
@@ -30,6 +30,8 @@ module ActiveShipping
             package_weight, package_value = 0, 0
             state = :filling_package
           when :filling_package
+            validate_package_quantity(packages.count)
+
             items_to_pack.each do |item|
               quantity = determine_fillable_quantity_for_package(item, maximum_weight, package_weight)
               package_weight += item_weight(quantity, item[:grams])
@@ -61,6 +63,10 @@ module ActiveShipping
 
           raise_excess_quantity_error if maybe_excess_package_quantity?(total_weight, maximum_weight)
         end
+      end
+
+      def validate_package_quantity(number_of_packages)
+        raise_excess_quantity_error if number_of_packages >= EXCESS_PACKAGE_QUANTITY_THRESHOLD
       end
 
       def raise_excess_quantity_error
--- a/test/unit/shipment_packer_test.rb
+++ b/test/unit/shipment_packer_test.rb
@@ -146,10 +146,35 @@ class ShipmentPackerTest < ActiveSupport::TestCase
     assert_equal 100, package.value
   end
 
-  test "excess packages" do
+  test "excess packages raised over threshold before packing begins" do
+    ActiveShipping::Package.expects(:new).never
+    items = [{ grams: 1, quantity: ShipmentPacker::EXCESS_PACKAGE_QUANTITY_THRESHOLD + 1, price: 1.0 }]
+
     assert_raises(ShipmentPacker::ExcessPackageQuantity) do
-      items = [{ grams: 1, quantity: ShipmentPacker::EXCESS_PACKAGE_QUANTITY_THRESHOLD + 1, price: 1.0 }]
       ShipmentPacker.pack(items, @dimensions, 1, 'USD')
+    end
+  end
+
+  test "excess packages not raised at threshold" do
+    items = [{ grams: 1, quantity: ShipmentPacker::EXCESS_PACKAGE_QUANTITY_THRESHOLD, price: 1.0 }]
+    packages = ShipmentPacker.pack(items, @dimensions, 1, 'USD')
+
+    assert_predicate packages, :present?
+  end
+
+  test "excess packages not raised below threshold" do
+    items = [{ grams: 1, quantity: ShipmentPacker::EXCESS_PACKAGE_QUANTITY_THRESHOLD - 1, price: 1.0 }]
+    packages = ShipmentPacker.pack(items, @dimensions, 1, 'USD')
+
+    assert_predicate packages, :present?
+  end
+
+  test "excess packages with slightly larger max weight than item weight" do
+    max_weight = 750
+    items = [{ grams: 500, quantity: ShipmentPacker::EXCESS_PACKAGE_QUANTITY_THRESHOLD + 1, price: 1.0 }]
+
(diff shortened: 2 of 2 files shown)


lib/active_shipping/shipment_packer.rb
```diff
@@ -20,7 +20,7 @@ def pack(items, dimensions, maximum_weight, currency)
 
         # Naive in that it assumes weight is equally distributed across all items
         # Should raise early enough in most cases
-        total_weight = calculate_total_weight(items, maximum_weight)
+        validate_total_weight(items, maximum_weight)
```
MalazAlamir: I realized that we dont actually do anything with the accumulator.. we only use it to validate the total weight, changed the method name and no longer using `inject`

test/unit/shipment_packer_test.rb
```diff
@@ -146,13 +146,37 @@ def test_cast_quantity_and_grams_to_int
...
+  def test_excess_packages_raised_over_threshold
     assert_raises(ShipmentPacker::ExcessPackageQuantity) do
-      items = [{:grams => 1, :quantity => ShipmentPacker::EXCESS_PACKAGE_QUANTITY_THRESHOLD + 1, :price => 1.0}]
+      items = [{ grams: 1, quantity: ShipmentPacker::EXCESS_PACKAGE_QUANTITY_THRESHOLD + 1, price: 1.0 }]
       ShipmentPacker.pack(items, @dimensions, 1, 'USD')
     end
   end
 
+  def test_excess_packages_not_raised_at_threshold
+    assert_nothing_raised do
```
jonathankwok: You'll never need an `assert_nothing_raised`, that's what a test inherently does. Any test fails if something is raised.

lib/active_shipping/shipment_packer.rb
```diff
@@ -50,27 +51,32 @@ def pack(items, dimensions, maximum_weight, currency)
...
+      def raise_excess_quantity_error
+        raise ExcessPackageQuantity, "Unable to pack more than #{EXCESS_PACKAGE_QUANTITY_THRESHOLD} packages"
+      end
+
       def overweight_item?(grams, maximum_weight)
         grams.to_i > maximum_weight
       end
 
-      def excess_package_quantity?(total_weight, maximum_weight)
+      def likely_excess_quantity?(total_weight, maximum_weight)
```
justinkrol: Not sure about the name, since if this returns `true`, it's not _likely_ an excess quantity, it _is_ an excess quantity, correct?
It needs to communicate that even if it returns `false`, it could still be an excess quantity, but I feel like this name communicates the opposite, that even if it returns `true`, it might _not_ be an excess quantity.

test/unit/shipment_packer_test.rb
```diff
@@ -146,13 +146,38 @@ def test_cast_quantity_and_grams_to_int
     assert_equal 100, package.value
   end
 
-  def test_excess_packages
+  def test_excess_packages_raised_over_threshold_before_packing_begins
```
qq99: can you change these to `test "etc etc etc" do` while you're here?

lib/active_shipping/shipment_packer.rb
qq99: `maybe`?

lib/active_shipping/shipment_packer.rb
```diff
@@ -40,6 +40,7 @@ def pack(items, dimensions, maximum_weight, currency)
             items_to_pack.reject! { |i| i[:quantity].to_i == 0 }
             state = :package_full
           when :package_full
+            validate_package_quantity(packages.count)
```
qq99: Is this valuable? If we get to this case, then the loop will terminate, and validating the quantity in general I figure is/was a way to put an upper bound on load limit on runtime.  When `:package_full` we will be doing no more work.  Unless we're concerned we may generate too big of a JSON payload?

Maybe we should be doing this during `:filling_package` instead?

I _think_ I might like to see this change as its own PR, since the bulk of this PR is renaming and cleaning up code, while this is introducing a new behaviour

MalazAlamir: hmm thats a good point, this can definitely be in `:filling_package`, as long as we're enforcing the upper bound at some point in the loop I'm okay with that.

I'll split this PR up

lib/active_shipping/shipment_packer.rb
justinkrol: I think `maybe` still has the same problem. I'm thinking something with a similar meaning to `for_sure_excess_quantity?` 
i.e. if it returns `true`, then it's _for sure_ an excess quantity
      if it returns `false`, then it's _not for sure_ and excess quantity

test/unit/shipment_packer_test.rb
```diff
@@ -146,13 +146,38 @@ def setup
...
+
+    assert_predicate packages, :present?
+  end
+
+  test "excess packages not raised below threshold" do
+    items = [{ grams: 1, quantity: ShipmentPacker::EXCESS_PACKAGE_QUANTITY_THRESHOLD - 1, price: 1.0 }]
+    packages = ShipmentPacker.pack(items, @dimensions, 1, 'USD')
+
+    assert_predicate packages, :present?
+  end
```
qq99: maybe 1 more test for grams=1, `quantity: ShipmentPacker::EXCESS_PACKAGE_QUANTITY_THRESHOLD +1`, and `max_weight=1`?

not sure if the other test below covers that (at least it's not as straightforward to read

MalazAlamir: already covered in https://github.com/Shopify/active_shipping/pull/484/files#diff-e550062946d31e0113e46265e32fe3b8R149

PR #487, opened 2017-02-28
Delivery estimates returning with USPS rates

developingchris:
To make the api of the USPS carrier consistent with UPs and FedEx as
much as possible, return the delivery_range with the rate response.

So as to not complicated existing workflows, I've used an option for the
optional node that must be returned to make commitments return.

### DIFF
--- a/test/remote/usps_test.rb
+++ b/test/remote/usps_test.rb
@@ -229,7 +229,7 @@ class RemoteUSPSTest < ActiveSupport::TestCase
     response = @carrier.find_rates(
        location_fixtures[:beverly_hills],
        location_fixtures[:new_york],
-      Package.new(16, [12, 6, 2], :units => :imperial),
+      Package.new(16, [12, 6, 2], units: :imperial),
       ship_date: Time.now
     )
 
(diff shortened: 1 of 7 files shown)


test/remote/usps_test.rb
```diff
@@ -225,6 +225,19 @@ def test_must_provide_login_creds_when_instantiating
     end
   end
 
+  def test_delivery_range_when_providing_ship_date
+    response = @carrier.find_rates(
+       location_fixtures[:beverly_hills],
+       location_fixtures[:new_york],
+      Package.new(16, [12, 6, 2], :units => :imperial),
```
developingchris: I know that colon syntax is a common critique. It felt weird to mix it in this test file.
So maybe that's my next pull request to clean up this test file.

developingchris:
@kmcphillips  This appears to be a travis configuration issue that was failing while S3 was having a major outage. What is the best way to re-run the CI and correct the lack of binary rubies in the setup before the tests get a chance to run?

developingchris:
going to revise this slightly.

PR #489, opened 2017-03-08
Strip 9 digit origin zip code for world rate requests

anthonyn60:
## What

This PR truncates origin +4 zip codes for USPS world rate requests using the `strip_zip` method.

## Why

`build_world_rate_request` may throw an invalid zip code error if a +4 origin zip code is used in its original form. `build_us_rate_request` gets around this by calling `strip_zip` on any zip codes passed into the XML for cleansing. I'm adding the same behaviour to the `build_world_rate_request` method when it creates its own `xml`.

### DIFF
--- a/test/unit/carriers/usps_test.rb
+++ b/test/unit/carriers/usps_test.rb
@@ -386,6 +386,13 @@ class USPSTest < ActiveSupport::TestCase
     assert request =~ /\>12345\</
   end
 
+  def test_strip_9_digit_zip_codes_world_rates
+    request = URI.decode(@carrier.send(:build_world_rate_request, location_fixtures[:beverly_hills_9_zip],
+                                        package_fixtures[:book], location_fixtures[:auckland], {}))
+    refute_match /\<OriginZip\>90210-1234/, request
+    assert_match /\<OriginZip\>90210/, request
+  end
+
   def test_maximum_weight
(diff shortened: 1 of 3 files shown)


anthonyn60:
Here's an example of an error I get when requesting an international rate with a +4 zip code:
 
screen shot 2017-03-06 at 11 50 19 am

test/unit/carriers/usps_test.rb
```diff
@@ -386,6 +386,13 @@ def test_strip_9_digit_zip_codes
     assert request =~ /\>12345\</
   end
 
+  def test_strip_9_digit_zip_codes_world_rates
+    request = URI.decode(@carrier.send(:build_world_rate_request, location_fixtures[:beverly_hills_9_zip],
+                                        package_fixtures[:book], location_fixtures[:auckland], {}))
+    assert !(request =~ /\>90210-1234\</)
```
jonathankwok: I recommend using http://apidock.com/ruby/MiniTest/Assertions/assert_matchhttp://apidock.com/ruby/MiniTest/Assertions/assert_match, not the generic assert.

jonathankwok: Also, rather than just the `\>`, it would be clearer to represent the whole tag, e.g. `/\<OriginZip\>90210/`

anthonyn60: Refactored to address your feedback. I agree, there's more clarity when the entire tag is included.

PR #494, opened 2017-03-13
Patch UPS to use old Kosovo country code

anthonyn60:
## What
Patch the `CountryCode` on a UPS XML request payload, changing the country code from `XK` to `KV` if necessary (Kosovo address).

## Why
In `ActiveUtils`, Kosovo's country code was recently updated from `KV` to `XK` (https://github.com/Shopify/active_utils/pull/79). UPS still uses Kosovo's old country code (`KV`), and will throw an invalid address if the new country code is used.

### DIFF
--- a/lib/active_shipping/carriers/ups.rb
+++ b/lib/active_shipping/carriers/ups.rb
@@ -127,6 +127,10 @@ module ActiveShipping
 
     IMPERIAL_COUNTRIES = %w(US LR MM)
 
+    COUNTRY_MAPPING = {
+      'XK' => 'KV'
+    }.freeze
+
     DEFAULT_SERVICE_NAME_TO_CODE = Hash[UPS::DEFAULT_SERVICES.to_a.map(&:reverse)]
     DEFAULT_SERVICE_NAME_TO_CODE['UPS 2nd Day Air'] = "02"
     DEFAULT_SERVICE_NAME_TO_CODE['UPS 3 Day Select'] = "12"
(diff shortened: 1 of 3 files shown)


.byebug_history
```diff
@@ -0,0 +1,2 @@
+quit
```
lucasuyezu: 👀

anthonyn60: Removed that

anthonyn60:
@lucasuyezu Added additional test coverage.

lib/active_shipping/carriers/ups.rb
```diff
@@ -1201,5 +1201,9 @@ def package_level_delivery_confirmation?(origin, destination)
       origin.country_code == destination.country_code ||
       [['US','PR'], ['PR','US']].include?([origin,destination].map(&:country_code))
     end
+
+    def patch_country_code(country_code)
```
mdking: This really doesn't communicate what it does. `patch` is so generic, can we just call this `mapped_country_code` or something? #NamingIsHard

e.g.

```
COUNTRY_MAPPING = {
  'XK' => 'KV'
}.freeze

...

def mapped_country_code(country_code)
{
  COUNTRY_MAPPING[country_code].presence || country_code
}
```

anthonyn60: I like this, makes countries easier to ~~patch~~ map in the future should we run into this again. 

Naming was tough, went back and forth with @jonathankwok on that one but I do think `map` describes the action best.

PR #496, opened 2017-03-14
add option to not include tax in rates for CP PWS

imran-iq:
Sometimes tax is not desired in the rates lookup (eg, you are calculating your own tax)
Add an option to exclude tax from rate lookup. Also fixed the test fixture as the numbers were not lining up.

### DIFF
--- a/lib/active_shipping/carriers/canada_post_pws.rb
+++ b/lib/active_shipping/carriers/canada_post_pws.rb
@@ -78,7 +78,7 @@ module ActiveShipping
       url = endpoint + "rs/ship/price"
       request  = build_rates_request(origin, destination, line_items, options, package, services)
       response = ssl_post(url, request, headers(options, RATE_MIMETYPE, RATE_MIMETYPE))
-      parse_rates_response(response, origin, destination)
+      parse_rates_response(response, origin, destination, !!options[:exclude_tax])
     rescue ActiveUtils::ResponseError, ActiveShipping::ResponseError => e
       error_response(e.response.body, CPPWSRateResponse)
     end
@@ -273,7 +273,7 @@ module ActiveShipping
       builder.to_xml
     end
 
-    def parse_rates_response(response, origin, destination)
+    def parse_rates_response(response, origin, destination, exclude_tax)
       doc = Nokogiri.XML(response)
       doc.remove_namespaces!
       raise ActiveShipping::ResponseError, "No Quotes" unless doc.at('price-quotes')
@@ -281,7 +281,7 @@ module ActiveShipping
       rates = doc.root.xpath('price-quote').map do |node|
         service_name  = node.at("service-name").text
         service_code  = node.at("service-code").text
-        total_price   = node.at('price-details/due').text
+        total_price   = price_from_node(node, exclude_tax)
         expected_date = expected_date_from_node(node)
         options = {
           :service_code   => service_code,
@@ -292,6 +292,14 @@ module ActiveShipping
         RateEstimate.new(origin, destination, @@name, service_name, options)
       end
       CPPWSRateResponse.new(true, "", {}, :rates => rates)
+    end
+
+    def price_from_node(node, exclude_tax)
+      price = node.at('price-details/due').text
+      return price unless exclude_tax
+      children = node.at('price-details/taxes').children
+      tax_total_cents = children.sum { |node| node.elem? ? Package.cents_from(node.text) : 0 }
+      Package.cents_from(price) - tax_total_cents
     end
 
     # tracking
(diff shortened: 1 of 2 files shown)


lib/active_shipping/carriers/canada_post_pws.rb
```diff
@@ -281,7 +282,11 @@ def parse_rates_response(response, origin, destination)
       rates = doc.root.xpath('price-quote').map do |node|
         service_name  = node.at("service-name").text
         service_code  = node.at("service-code").text
-        total_price   = node.at('price-details/due').text
+        total_price   = if @rates_exclude_tax
```
nicolaslupien: You can use string template to do it in a single line and still readable
`total_price = "price-details/#{@rates_exclude_tax ? 'base' : 'due' }"`

imran-iq:
This wont work, fuel surchage is included in tax, reworking.

lib/active_shipping/carriers/canada_post_pws.rb
```diff
@@ -294,6 +294,16 @@ def parse_rates_response(response, origin, destination)
       CPPWSRateResponse.new(true, "", {}, :rates => rates)
     end
 
+    def price_from_node(node, exclude_tax)
+      price = node.at('price-details/due').text
+      return price unless exclude_tax
+      taxes = 0
+      node.at('price-details/taxes').children.each do |node|
+        taxes += Package.cents_from(node.text) if node.elem?
```
imran-iq: From some reason it thinks `taxes` has 7 children (counts newline in the xml response as a child). These newlines are of type `Text` where as the actual tax xml fields (gst,pst,hst) are of type `Element`, using `node.elem?` I skip over the newlines.

nicolaslupien: nit: there's also a cleaner way to update `taxes` without doing a side effect in a block.

```ruby
children = node.at('price-details/taxes').children
elem_nodes = children.select { |node| node.elem? }
taxes = elem_nodes.sum { |node| Package.cents_from(node.text) }
```

imran-iq: While it is cleaner we end up iterating over the same list twice (though smaller second time) with my current code we only iterate once. But if we are in favour of not having the block i will implement the cleaner solution.

lib/active_shipping/carriers/canada_post_pws.rb
```diff
@@ -273,15 +273,15 @@ def build_rates_request(origin, destination, line_items = [], options = {}, pack
       builder.to_xml
     end
 
-    def parse_rates_response(response, origin, destination)
+    def parse_rates_response(response, origin, destination, exclude_tax=false)
```
jonathankwok: You don't need the optional default param, `!!nil` is `false`.

lib/active_shipping/carriers/canada_post_pws.rb
jonathankwok: Why not

```ruby
children = node.at('price-details/taxes').children
tax_total_cents = children.sum { |node| node.elem? ? Package.cents_from(node.text) : 0 }
```

I also renamed the variable `taxes` to `taxes_total_cents` because let's be clear in what the value it represents is.

test/unit/carriers/canada_post_pws_rating_test.rb
```diff
@@ -112,6 +112,25 @@ def test_find_rates_line_items_single_object
...
 
+  def test_find_rates_excluding_tax
+    response = xml_fixture('canadapost_pws/rates_info')
+    expected_headers = {
+      'Authorization'   => "#{@cp.send(:encoded_authorization)}",
+      'Accept-Language' => 'en-CA',
+      'Accept'          => 'application/vnd.cpc.ship.rate+xml',
+      'Content-Type'    => 'application/vnd.cpc.ship.rate+xml'
+    }
+    CanadaPostPWS.any_instance.expects(:ssl_post).with(anything, anything, expected_headers).returns(response).times(2)
```
jonathankwok: Style; `.twice == .times(2)`

lib/active_shipping/carriers/canada_post_pws.rb
```diff
@@ -294,6 +294,14 @@ def parse_rates_response(response, origin, destination)
       CPPWSRateResponse.new(true, "", {}, :rates => rates)
     end
 
+    def price_from_node(node, exclude_tax)
+      price = node.at('price-details/due').text
+      return price unless exclude_tax
+      children = node.at('price-details/taxes').children
+      tax_total_cents = children.sum { |node| node.elem? ? Package.cents_from(node.text) : 0 }
```
imran-iq: Re-adding explanation since it was squashed

From some reason it thinks taxes has 7 children (counts newline in the xml response as a child). These newlines are of type `Text` where as the actual tax xml fields (gst,pst,hst) are of type `Element`, using `node.elem?` I skip over the newlines.

PR #502, opened 2017-04-04
Bump nokogiri version to 1.7.1 due to security vulnerability

bibstha:
Reference: https://github.com/sparklemotion/nokogiri/issues/1615


MalazAlamir:
talked in person, will need to also bump active_support and active_utils to get the builds to pass. We might need to create a new branch if we want to use this downstream before the major bump

MalazAlamir:
Also see: https://github.com/Shopify/active_shipping/pull/500

Anyone on v1.13.2 isn't locked at 1.6.8 downstream anymore, they can use 1.7.1

bibstha:
Hi guys, I didn't realize the stable branch is 1.9-stable and not master.
@MalazAlamir told me 1.9-stable is kept so that it still supports ruby <2.1.0 while latest nokogiri supports only >= 2.1.0.

To keep up to date with the security fixes, active_shipping will eventually have to remove the support for older ruby versions.

Since we are not merging into master yet, shall we instead 

* create a 1.10-stable branch that supports ruby >=2.1.0
* phase out 1.9-stable (people can use it if they are ok with the security issues) but ask them to move to new ruby version
* eventually move things to master

@kmcphillips @qq99 thoughts?

qq99:
I can't really weigh in on the strategy, but it'd be nice to do away with 1-9-stable for good and use master w/ tags for 2.0.0+ release

PR #503, opened 2017-04-13
remove gsub! form canada_post_pws

imran-iq:
Removing `gsub!` from Canada post PWS as everything is referenced in Ruby this causes some issues where the location you created using `ActiveShipping::Location#from` ends up modifying the original object (which has many undesirable effects)

### DIFF
--- a/lib/active_shipping/carriers/canada_post_pws.rb
+++ b/lib/active_shipping/carriers/canada_post_pws.rb
@@ -359,9 +359,6 @@ module ActiveShipping
     # :cod, :cod_amount, :insurance, :insurance_amount, :signature_required, :pa18, :pa19, :hfp, :dns, :lad
     #
     def build_shipment_request(origin, destination, package, line_items = [], options = {})
-      origin = sanitize_location(origin)
-      destination = sanitize_location(destination)
-
       builder = Nokogiri::XML::Builder.new do |xml|
         xml.public_send('non-contract-shipment', :xmlns => "http://www.canadapost.ca/ws/ncshipment") do
           xml.public_send('delivery-spec') do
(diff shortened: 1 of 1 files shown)


lib/active_shipping/carriers/canada_post_pws.rb
```diff
@@ -776,17 +772,8 @@ def shipping_date(options)
...
-      Location.new(location_hash)
-    end
-
-    def sanitize_zip(hash)
-      [:postal_code, :zip].each do |attr|
-        hash[attr].gsub!(/\s+/, '') if hash[attr]
-      end
-      hash
+    def get_sanitized_postal_code(location)
+      location.try(:postal_code).gsub(/\s+/, '')
```
qq99: `undefined method `gsub' for nil:NilClass`
`.try(:gsub, /\s+/, '')`

lib/active_shipping/carriers/canada_post_pws.rb
```diff
@@ -776,17 +772,8 @@ def shipping_date(options)
       DateTime.strptime((options[:shipping_date] || Time.now).to_s, "%Y-%m-%d")
     end
 
-    def sanitize_location(location)
-      location_hash = location.is_a?(Location) ? location.to_hash : location
```
titanoboa: The old version looks like location could be an instance of `Location` or a hash. If it is a hash, the new version if I'm not mistaken returns `nil` since a hash does not respond to `postal_code`. Is this intended?

lib/active_shipping/carriers/canada_post_pws.rb
```diff
@@ -446,7 +445,7 @@ def references_node(xml, options)
     end
 
     def shipment_customs_node(xml, destination, line_items, options)
-      return unless destination.country_code != 'CA'
+      return unless location_from_hash(destination).country_code != 'CA'
```
qq99: this one is already a location?

imran-iq: nope, it used to be but i moved that code out

PR #512, opened 2017-05-11
Add new Correios service codes and update defaults

lucasuyezu:
### Summary
Same bug fixed on #510, but porting it back to the `1-9-stable` branch.

From the original description:
### Summary
Due to some API changes, the Correios' endpoint deprecated service codes 41106 and 40010, in favour of the new ones: '04510', and '04014'. This PR adds the new service codes and updates the list of default services.

### Observations
- The new service codes must be used as is, with the leading zero. If used in the same format as the pre-existing ones, the request will fail.
- I did not update the existing services to strings to avoid having downstream problems. Maybe we can do that in the next major release.

### DIFF
--- a/lib/active_shipping/carriers/correios.rb
+++ b/lib/active_shipping/carriers/correios.rb
@@ -21,8 +21,10 @@ module ActiveShipping
 
     protected
 
-    DEFAULT_SERVICES = [41106, 40010]
+    DEFAULT_SERVICES = ['04510', '04014']
     AVAILABLE_SERVICES = {
+      '04510' => 'PAC sem contrato',
+      '04014' => 'SEDEX sem contrato',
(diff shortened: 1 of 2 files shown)


lib/active_shipping/carriers/correios.rb
```diff
@@ -21,8 +21,10 @@ def self.available_services
 
     protected
 
-    DEFAULT_SERVICES = [41106, 40010]
+    DEFAULT_SERVICES = ['04510', '04014']
```
MalazAlamir: just curious - why strings now?

lucasuyezu: If we send `4510` instead of `04510` to the endpoint, it'll return an error saying the service is invalid.

PR #513, opened 2017-05-12
Make our remote UPS tests consider Saturday as a business day

lucasuyezu:
### Problem
Some of our remote UPS tests start failing on Thursday evening and keep failing throughout the remainder of the week because UPS considers Saturday as a business day.

### Solution
Add Saturday as a business day for these tests.

cc @mdking


bibstha:
Is this the reason why USPS tests failed today?
Also, whats the reason why they pass sometimes and they fail other times?

Issue #461, opened 2017-01-10
Depend on measured 2.0 or higher


kmcphillips:
We are working on the 2.0 milestone for `measured` and will have a full version released soon.
https://github.com/Shopify/measured/milestone/1

Version 2.0 of this gem should depend explicitly on 2.0 of measured. @benwah added it in #450. cc @thegedge

garethson:
Done.

kmcphillips:
👏

PR #450, opened 2016-12-20
Main commit getting rid of Quantified dependency - And replacing with Measured 1.6.0

benwah:
The changes are pretty simple - Everything still seems to work.

PR #521, opened 2017-05-24
Bump measured to 2.0

garethson:
Bumps [measured](https://github.com/Shopify/measured) to 2.0.

Diff https://github.com/Shopify/measured/compare/v1.6.0...v2.0.0

Addresses https://github.com/Shopify/active_shipping/issues/461.

### DIFF
--- a/lib/active_shipping/package.rb
+++ b/lib/active_shipping/package.rb
@@ -13,19 +13,16 @@ module ActiveShipping #:nodoc:
 
       @dimensions = [dimensions].flatten.reject(&:nil?)
 
-      imperial = (options[:units] == :imperial) ||
-                 ([grams_or_ounces, *dimensions].all? { |m| m.respond_to?(:unit) && m.unit.to_sym == :imperial })
+      imperial = (options[:units] == :imperial)
 
       weight_imperial = dimensions_imperial = imperial if options.include?(:units)
 
       if options.include?(:weight_units)
-        weight_imperial = (options[:weight_units] == :imperial) ||
-                          (grams_or_ounces.respond_to?(:unit) && m.unit.to_sym == :imperial)
+        weight_imperial = (options[:weight_units] == :imperial)
       end
 
       if options.include?(:dim_units)
-        dimensions_imperial = (options[:dim_units] == :imperial) ||
-                              (dimensions && dimensions.all? { |m| m.respond_to?(:unit) && m.unit.to_sym == :imperial })
+        dimensions_imperial = (options[:dim_units] == :imperial)
       end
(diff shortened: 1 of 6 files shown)


lib/active_shipping/package.rb
```diff
@@ -66,35 +66,35 @@ def gift?
...
     def pounds(options = {})
-      weight(options).convert_to(:lb).value
+      weight(options).convert_to(:lb).value.to_f
     end
     alias_method :lb, :pounds
     alias_method :lbs, :pounds
 
     def kilograms(options = {})
-      weight(options).convert_to(:kg).value
+      weight(options).convert_to(:kg).value.to_f
```
garethson: Measured now uses rationals if it can, so we need to `to_f` these.

lib/active_shipping/package.rb
```diff
@@ -13,19 +13,16 @@ def initialize(grams_or_ounces, dimensions, options = {})
 
       @dimensions = [dimensions].flatten.reject(&:nil?)
 
-      imperial = (options[:units] == :imperial) ||
-                 ([grams_or_ounces, *dimensions].all? { |m| m.respond_to?(:unit) && m.unit.to_sym == :imperial })
```
garethson: All these checks for `:unit` have been obsolete since [we remove quantified](https://github.com/Shopify/active_shipping/pull/450).

lib/active_shipping/package_item.rb
```diff
@@ -61,7 +60,7 @@ def kilograms(options = {})
 
     def attribute_from_metric_or_imperial(obj, klass, metric_unit, imperial_unit)
       if obj.is_a?(klass)
-        return value
+        return obj
```
garethson: @kmcphillips  This was the bug here that was causing the strange behavior in the test below.

kmcphillips: Oh great. Yeah I have some tests that I added after to specify behaviour that isn't used.

To be real, for 2.0 we should kill most of those code paths anyway.

kmcphillips: Also, both these `return` statements are unnecessary.

garethson:
@kmcphillips You're correct. Will re-ping when I clean this up.

kmcphillips:
Depending on how aggressively you want to refactor, you could write a measured extension class that understands `:metric` `:imperial` and remove the `(@unit_system ?` checks in a _ton_ of places.

garethson:
@kmcphillips  Cleaned up remaining usage of old Quantified methods and extraneous conversion methods.

PR #522, opened 2017-05-26
Bump master version gemspec.

garethson:
As we work towards a 2.0, let's set this version to what it really is.


lib/active_shipping/version.rb
```diff
@@ -1,3 +1,3 @@
 module ActiveShipping
-  VERSION = "1.9.2"
+  VERSION = "2.0.0.pre"
```
kmcphillips: `.pre1`

PR #523, opened 2017-05-26
Adhere to property order if using objects with hash access.

garethson:
[Recent changes](https://github.com/Shopify/active_shipping/pull/446/files#diff-808f0e47ee0b93d89184bb01b6d1f652L67) to the code to read attributes for locations resulted in a very subtle change in behaviour for objects that have both hash access and methods that return values, such as ActiveRecord objects.

Example: An `Address` model with both `province` column and `province_code` method. We used to return "CA" but we now return "California", and we've always been missing a test around this behaviour.

The root cause of this is an interface that is exceptionally wide, and subsequent version of this library we will move to a more simple and explicit interface.

### DIFF
--- a/lib/active_shipping/location.rb
+++ b/lib/active_shipping/location.rb
@@ -64,12 +64,7 @@ module ActiveShipping #:nodoc:
 
       attributes = {}
 
-      hash_access = begin
-        object[:some_symbol]
-        true
-      rescue
-        false
-      end
+      hash_access = object.respond_to?(:[])
 
       ATTRIBUTE_ALIASES.each do |attribute, aliases|
         aliases.detect do |sym|
--- a/lib/active_shipping/location.rb
+++ b/lib/active_shipping/location.rb
@@ -73,10 +73,11 @@ module ActiveShipping #:nodoc:
 
       ATTRIBUTE_ALIASES.each do |attribute, aliases|
         aliases.detect do |sym|
-          value = if hash_access
-            object[sym]
-          elsif object.respond_to?(sym)
-            object.send(sym)
+          value = object[sym] if hash_access
+          if !value &&
+            object.respond_to?(sym) &&
+            (!hash_access || !Hash.public_instance_methods.include?(sym))
+            value = object.send(sym)
           end
 
           attributes[attribute] = value if value
--- a/test/unit/location_test.rb
+++ b/test/unit/location_test.rb
@@ -47,7 +47,10 @@ class LocationTest < ActiveSupport::TestCase
       def method_missing(method)
         @hash[method]
       end
-      def respond_to?(method) ; true ; end
+      def respond_to?(method)
+        return false if method == :[]
+        true
+      end
     end.new(@attributes_hash)
 
     location = Location.from(object)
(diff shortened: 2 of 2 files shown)


lib/active_shipping/location.rb
```diff
@@ -73,10 +73,11 @@ def self.from(object, options = {})
 
       ATTRIBUTE_ALIASES.each do |attribute, aliases|
         aliases.detect do |sym|
-          value = if hash_access
```
garethson: So if you have an active record object with `province` and `province_code`, because it has `hash_access`, we were preferring `province` instead of `province_code`.

lib/active_shipping/location.rb
```diff
@@ -73,10 +73,11 @@ def self.from(object, options = {})
...
-            object[sym]
-          elsif object.respond_to?(sym)
-            object.send(sym)
+          value = object[sym] if hash_access
+          if !value &&
```
garethson: Unfortunately complicated, do to how loosely we alloy data to be passed into this class.

lib/active_shipping/location.rb
```diff
@@ -73,10 +73,11 @@ def self.from(object, options = {})
...
+            object.respond_to?(sym) &&
+            (!hash_access || !Hash.public_instance_methods.include?(sym))
+            value = object.send(sym)
           end
```
benwah: I'm wondering if we organize the code a little simpler by extracting the `hash_access` and `!hash_access`
behaviour in their own class methods, perhaps something like this?

Also *second point*:

Seems like `respond_to?` only checks for public_methods, so we could safely use `public_send` here rather then `send`.

```ruby
    class << self
      def from(object, options = {})
        return object if object.is_a?(ActiveShipping::Location)

        attributes = {}

        accessor = if object.respond_to?(:[])
          method(:hash_accessor)
        else
          method(:public_method_accessor)
        end

        ATTRIBUTE_ALIASES.each do |attribute, aliases|
          aliases.detect do |sym|
            value = accessor.(object, sym)
            attributes[attribute] = value if value
          end
        end

        attributes.delete(:address_type) unless ADDRESS_TYPES.include?(attributes[:address_type].to_s)

        new(attributes.update(options))
      end

      def hash_accessor(object, sym)
        value = object[sym]

        if (
          !value &&
          object.respond_to?(sym) &&
          !Hash.public_instance_methods.include?(sym)
        )
          value = public_method_accessor(sym)
        end

        value
      end

      def public_method_accessor(object, sym)
        object.public_send(sym)
      end
```

test/unit/location_test.rb
```diff
@@ -63,6 +63,20 @@ def respond_to?(method) ; true ; end
     assert_equal @attributes_hash[:name], location.name
   end
 
+  test ".from adheres to propery order even if hash access is available" do
+    object = Class.new do
+      def [](_index)
+        return "California" if index == :province
+        nil
```
benwah: Nice has implementation ;)

lib/active_shipping/location.rb
benwah: This is just a suggestion - Perhaps not a great one, so take with a grain of salt..

test/unit/location_test.rb
benwah: BTW - I think you gotta rename `_index` to `index`

lib/active_shipping/location.rb
garethson: I'll give this a shot, thanks!

garethson:
@benwah Cleaned up using `respond_to`, much better. Had to tweak an older test that was making bad assumptions around `respond_to`. 👍

Issue #516, opened 2017-05-17
Catch ActiveUtils::ResponseError during CanadaPostPWS receipt request


bibstha:
Right now we do not trap the ActiveUtils::ResponseError [here](https://github.com/Shopify/active_shipping/blob/v1.14.0/lib/active_shipping/carriers/canada_post_pws.rb#L115).
We should trap this and return an ActiveShipping::ResponseError instead.

anthonyn60:
This one is a little tricky just because `parse_shipment_receipt_response` doesn't return a `CPPWS` response but rather a simple `Hash` so we can't use `error_response` like the other rescue statements.

PR #524, opened 2017-05-30
Trap ActiveUtils::ResponseError and return ActiveShipping::ResponseError instead

anthonyn60:
### WHAT

Currently when we use `find_shipment_receipt` on Canada Post, in the case where the receipt is not found, `ssl_get` returns an `ActiveUtils::ResponseError`. This PR rescues that error and returns an `ActiveShipping::ResponseError` instead.

### WHY

We should only be raising `ActiveShipping` errors in this gem.

With most carriers, we rescue bad API responses and build a corresponding variation of the `ActiveShipping::Response` object to return so that the method fails gracefully. 

The Canada Post receipt is tricky because when it's parsed, it doesn't return a `Response` object, rather it's just a hash. Rescuing and returning a `Response` object here doesn't make sense so I'm making sure it at least raises an `ActiveShipping` related error.

Closes https://github.com/Shopify/active_shipping/issues/516

### DIFF
--- a/lib/active_shipping/carriers/canada_post_pws.rb
+++ b/lib/active_shipping/carriers/canada_post_pws.rb
@@ -115,6 +115,8 @@ module ActiveShipping
     def find_shipment_receipt(shipping_id, options = {})
       response = ssl_get(shipment_receipt_url(shipping_id, options), headers(options, SHIPMENT_MIMETYPE, SHIPMENT_MIMETYPE))
       parse_shipment_receipt_response(response)
+    rescue ActiveUtils::ResponseError => e
+      raise ActiveShipping::ResponseError.new(e.message)
     end
 
     def retrieve_shipping_label(shipping_response, options = {})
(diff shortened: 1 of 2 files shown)


garethson:
@anthonyn60 Because i'm not clear myself, I don't see us actually rescuing `ActiveUtils::ResponseError` in many of our other carriers. Why do we not want to allow `ActiveUtils::ResponseError`'s out of this library?

jonathankwok:
> We should only be raising `ActiveShipping` errors in this gem.

Not arguing against this but I'm interested in hearing an argument for it.

anthonyn60:
@garethson It's currently being rescued in Stamps, Shipwire, Australia Post, and Canada Post. Other carriers like Correios and New Zealand post have catch-all `rescue` statements that return an `ActiveShipping` specific response.

We also have patterns of rescuing non-`ActiveShipping` related errors and raising `ActiveShipping` errors in carriers like [Fedex](https://github.com/Shopify/active_shipping/blob/master/lib/active_shipping/carriers/fedex.rb#L789) and [UPS](https://github.com/Shopify/active_shipping/blob/master/lib/active_shipping/carriers/ups.rb#L810) where we rescue Nokogiri errors. I like the pattern because if I'm writing an app which uses a method within the `ActiveShipping` gem, I don't want explicit awareness of dependencies in my code.

I've updated the PR description to provide more context.

CC: @jonathankwok

garethson:
🆒  That makes sense to me.

PR #530, opened 2017-07-21
Support email field on Location

tamcgoey:
As more international carriers are supported with `ActiveShipping`, we're needing to support more contact methods outside of just the current implemented fields: `phone` & `fax`.

This PR adds support for a `email` field on `Location`, this way email addresses can be passed on to a carrier.


tamcgoey:
CI is failing do to a flaky UPS remote test. Should be save to deploy still.

lib/active_shipping/location.rb
```diff
@@ -13,6 +13,7 @@ class Location
       address3: [:address3],
       phone: [:phone, :phone_number],
       fax: [:fax, :fax_number],
+      email: [:email, :email_address],
```
garethson: Do you really need `email_address` here? These attribute aliases lead to code that's hard to determine what's actually being passed in / used.

tamcgoey: Thought it would be useful in the library, but I think it's okay to drop it in this case.

tamcgoey:
@garethson dropped the alias for `email`.

PR #532, opened 2017-07-27
fix Location#== with nil

joevandyk:
`location == nil` will blow up. This fixes that.

(We had code like `location != nil` for some reason in our codebase)


joevandyk:
@tamcgoey @garethson any chance this could get reviewed?

lib/active_shipping/location.rb
```diff
@@ -160,6 +160,7 @@ def address2_and_3
     end
 
     def ==(other)
+      return false if !other.respond_to?(:to_hash)
```
javierhonduco: I'd go for something like
```patch
     def ==(other)
-      to_hash == other.to_hash
+      to_hash == other&.to_hash
     end
```

test/unit/location_test.rb
```diff
@@ -231,4 +231,8 @@ def province_code
     assert_equal @location, another_location
     refute_equal @location, Location.new({})
   end
+
+  test "#== works with nil" do
+    refute_equal @location, nil
```
javierhonduco: `refute_nil` could be used here

lib/active_shipping/location.rb
joevandyk: That’s a newer ruby feature.. is that ok to do?

joevandyk: I think `&.` was introduced in Ruby 2.3 and Travis is still testing 2.2.

javierhonduco: Not sure about that! Ruby 2.3 was introduced 2 years ago, but probably Thomas has an answer :)

test/unit/location_test.rb
javierhonduco: My bad, using `refute_nil` is not ok, as we don't want to check that the value is not `nil` (which I think should not happen), but that the comparison with `nil` returns `false` 😅

lib/active_shipping/location.rb
tamcgoey: Yea, we wouldn't be able to use the safe navigator operator here, since `ActiveShipping` is still supporting 2.2.

tamcgoey:
👋  @joevandyk, sorry for the delay getting to this PR. As of [v2.0 we're not actively maintaining](https://github.com/Shopify/active_shipping#versions) this gem anymore, but we're always looking for maintainers.

Issue #533, opened 2017-08-08
can you clarify 1.x vs 2.x?


gilesbowkett:
The readme for this repo says:

> Note: 2.x contains breaking changes, please see the changelog. Shopify will not be actively contibuting to the 2.0 version of this gem and is looking for maintainers. We have released 1.9 and will only backport small fixes to this version, on branch 1-9-stable, and they should be on master first.

However, the most recent 1.x version seems to be 1.14, so 1.9 was a while ago. And two recent contributors to the changelog on `master`, @jonathankwok and @tamcgoey, seem to work at Shopify (I took a look at your GitHub bios). So it seems as if Shopify is supporting 2.0+ after all.

Is this what's going on? Maybe Shopify decided not to support 2.x, then changed its mind, so to speak?

Sorry to be a pain, I just want to make sure I'm using the right version of the gem.

gilesbowkett:
update: I figured 2.x is supported now, so I submitted a small PR which removed the part of the `README` which confused me.

tamcgoey:
👋  @gilesbowkett thanks for reaching out regarding the README. We aren't actively maintaining ActiveShipping moving forward despite having a couple of bumps recently.

I'll update the README to make that clear.

As for 1.x release, this code for the most part lives in `1-9-stable` branch, hence why it was originally noted to only stop at `1.9`. If you're looking for the 1.x releases, it's probably best to check out our [tagged releases](https://github.com/Shopify/active_shipping/tags).

tamcgoey:
Follow up PR - #535

gilesbowkett:
@tamcgoey thanks! so you're not actively maintaining it at all? or you're just not maintaining 2.x? from the followup PR it sounds like you're just not maintaining 2.x. if so, good news. 😁

tamcgoey:
Only fixes that we need to apply will done (e.g. for security reasons), but as for any on going development we're no longer going to contributing those to ActiveShipping.

gilesbowkett:
@tamcgoey thanks, good to know.

tamcgoey:
Closing this, as #535 has been merged.

PR #534, opened 2017-08-08
removed 1.x vs 2.x versions disclaimer

gilesbowkett:
this doesn't seem to be true any more, so I removed it.

related: https://github.com/Shopify/active_shipping/issues/533


tamcgoey:
Updated the README to better clarify our involvement and `1.x` releases.

Addressed with PR #535.

PR #538, opened 2017-09-07
Fix CanadaPostPWS#find_rates not sending dimensions

maartenvg:
### Problem

Currently the common interface for `find_rates` is `def find_rates(origin, destination, packages, options = {})`. This is the case for all carriers, except CanadaPost and CanadaPostPWS. They use `def find_rates(origin, destination, line_items = [], options = {}, package = nil, services = [])` where `line_items` supposedly is an array of `PackageItems`. This then bubbles down via `build_rate_request` to `parcel_node`. There it takes `package` for the dimensions and package or line items for the weight. 

Because the interface for CanadaPostPWS deviates from the others, any user that wants to send a consistent set of params to the carriers ends up sending `packages` to something that is called `line_items` on canada post, and leaves `package` `nil`. Which then results in a `parcel_node` with no package and only line items. This results in a parcel XML node that only contains the weight (from the sum of the line items) and no dimensions.

### Solution

In this PR I'm changing a few things at once 
- Use same method signature as all the other `find_rates` endpoints: i.e. got rid of `line_items` and replaced it with `packages`. I left the `services` array for now.
- Since CP only supports one package per shipment, I then grab the first package to pass into `build_rates_request`
- With only one package and no line items `parcel_node` changes a bit to always get its information from that one package.


maartenvg:
Closing since this is going to be solved elsewhere (outside of this repo)

PR #550, opened 2018-01-17
Support Active Support 5.2

sgrif:
cc @Shopify/rails

### DIFF
--- a/active_shipping.gemspec
+++ b/active_shipping.gemspec
@@ -21,7 +21,7 @@ Gem::Specification.new do |s|
 
 
   s.add_dependency("measured", ">= 2.0")
-  s.add_dependency("activesupport", ">= 4.2", "< 5.2.0")
+  s.add_dependency("activesupport", ">= 4.2", "<= 5.2.0.beta2")
   s.add_dependency("active_utils", "~> 3.3.1")
   s.add_dependency("nokogiri", ">= 1.6")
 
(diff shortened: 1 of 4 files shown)


active_shipping.gemspec
```diff
@@ -21,7 +21,7 @@ Gem::Specification.new do |s|
 
 
   s.add_dependency("measured", ">= 2.0")
-  s.add_dependency("activesupport", ">= 4.2", "< 5.2.0")
+  s.add_dependency("activesupport", ">= 4.2", "<= 5.2.0.beta2")
```
rafaelfranca: Maybe it is better to remove the upper restriction like we are doing in the projects in the rails organization.

rafaelfranca: BTW, beta2 is already allowed by `< 5.2.0` since for rubygems any pre release is less than the final release.

Edouard-chin:
Last previous broken build had 4 failures on 5.0, there is 5 failures on 5.1 and 5.2, the  extra one is `test_domestic_combined_response` and doesn't seems to be related so LGTM.

Edouard-chin (approved):

active_shipping.gemspec
sgrif: I'm not fond of that policy personally, and yeah I realized this is mostly pointless until 5.2 final is out

lbrito1:
Hi,
Even though this moves forward, won't we need to update [active_utils](https://github.com/shopify/active_utils) as well? Currently it depends on activesupport < 5.2.0.

Edouard-chin:
#556

PR #556, opened 2018-04-07
Fix failing tests

michaelsunglee:
This PR fixes failing tests for Kunaki, USPS, and Correios.
**Kunaki**
- Change the URL scheme
- Replace old SKUs with new test SKUs provided by Kunaki

**USPS**
- Removed the outdated test `test_tracking_with_attempted_delivery `. This test used a live tracking number which is now out of use and thus returned an error. Adding a new "live" tracking number is only temporary and isn't a solution. USPS does not provide any testing tracking numbers either. 

**Correios**
- It seems that this carrier has stopped supporting their API endpoints. All sources point to this sign. For example [this](http://correiosapi.apphb.com/) site refers to [this](http://correiosapi.apphb.com/cep/76873274) simple zip lookup which fails. Other open source wrappers for this API have either reported this problem or have been archived themselves (for [example](https://github.com/emersonsoares/address-api)).  I've tested a few other wrappers and they all seem to fail. For these reasons I think Correios should not be supported (at least for now) in active_shipping as the failing tests are slowing development.

This PR also fixes the Travis CI build errors around activesupport (it required Ruby version >= 2.4.1). The build also supports activesupport 6.0.0 and so the dependencies reflect that (as well as creating activesupport51.gemfile for more test builds).

Please review changes. @rafaelfranca  @Edouard-chin @tamcgoey @garethson


.travis.yml
```diff
@@ -5,16 +5,19 @@ sudo: false
...
 - "2.2.2"
 - "2.3.1"
-- "2.4.0"
+- "2.4.1"
 
 gemfile:
 - gemfiles/activesupport42.gemfile
 - gemfiles/activesupport50.gemfile
-- gemfiles/activesupport_master.gemfile
+- gemfiles/activesupport51.gemfile
```
rafaelfranca: We are missing tests with 5.2.

gemfiles/activesupport_master.gemfile
```diff
@@ -2,4 +2,4 @@ source "https://rubygems.org"
 
 gemspec path: '..'
 
-gem 'activesupport', github: 'rails/rails'
+gem 'activesupport', github: 'rails/rails', :submodules => true
```
rafaelfranca: why submodules? I don't think rails has submodules in the repository.

rafaelfranca:
> Correios

As a Brazilian it makes me sad that correios broke their API 😢

.travis.yml
michaelsunglee: Added a gemfile for activesupport 5.2

PR #557, opened 2018-04-09
Canada Post Contract shipping

bperkinson:
Canada Post has 2 distinct account types (small business & commercial). Each have a distinct way of creating shipments.  The current implementation only allowed for small business accounts to create shipments.

Added functionality for Canada Post commercial shipments (using contract #)  
Added functionality to create shipment with a bundled return label (commercial only function)
Fixed remote tests for non contract shipments
Added remote and unit tests for contract shipments


bperkinson:
The Canada Post test fail because they are using a enterprise account trying to ship using the small business services.

How can I change the accounts used for running the automated tests since they are in environment variables stored on the test machine?

test account for enterprise / contract shipping - https://www.canadapost.ca/cpo/mc/business/productsservices/developers/services/fundamentals.jsf#TestValues

There is no open world account for small business.  It is however possible to sign up for one.

